PR #25, opened 2026-07-15
feat(moe): --lock-dense pins the dense weights against reclaim

Helldez:
## Why

The route trace ([bench-data/2026-07-15](docs/bench-data/2026-07-15-route-trace/)) showed gpt-oss-120b's major faults growing **267 → 8431 per token** during a run, with `compute_ms` tracking them **539 → 1767 ms**. Warming the dense weights only fills the page cache, which is reclaimable by definition — the streamer's own expert reads evict them right back out, and they re-fault from flash *inside* the next decode, where the wait is billed to "compute".

## What

`--lock-dense` mlocks the dense ranges (the complement of the expert ranges — the same set `warm_dense` already sweeps) so the kernel cannot take them back.

- **Not** llama.cpp's `mparams.use_mlock`: that pins the whole mapping, experts included, which would try to pin 58 GiB of gpt-oss onto an 11 GiB phone. Only the non-expert complement is locked; the expert region must stay reclaimable.
- The mapping base is derived from the captured expert tensors (they point into the model's mmap until the rebind), validated for consistency across every captured tensor — if they disagree it is not one contiguous mapping and locking is skipped rather than guessed.
- Ordering is load-bearing: the lock runs **before** the rebind (which moves the tensors off the mmap) and **before** the auto-budget probe (pinned pages leave `MemAvailable`).
- New `pio::vm_lock` / `vm_unlock` / `vm_lock_limit_raise` primitives, both platforms.

## Off by default — and why

The RAM has to come from somewhere, and on an auto budget it comes out of the expert cache. **That trade is already measured in this repo** ([bench-data/2026-07-14-warmup](docs/bench-data/2026-07-14-warmup/)): reserving Gemma's dense set cost budget 4000→2909 MiB, hit 82.9→73.5 % and **tok/s 4.76→3.76**. Locking reproduces that arithmetic exactly, so a default-on flag would be a ~21 % regression on the models that fit a useful cache.

It is a **>RAM knob**: worth it where hit rate is cheap to sell (gpt-oss holds ~5 % of a 58 GiB expert set and faults thousands of pages/token), not on Qwen/Gemma which fault 18–49 times/token and have nothing to gain. Judge it on `majflt/tok`, not on principle.

Best-effort by contract: `RLIMIT_MEMLOCK` caps what a process may pin, so `vm_lock` chunks the range, reports what it pinned, and the remainder degrades to warm-only.

## Verification

- Host build clean, **all 4 byte-identity gates pass** (no data-path change).
- Driven end-to-end on both gate models: lock fires and pins (`bmoe: dense lock — N MiB pinned`), `--lock-dense` off by default, warm-off/lock-on works (mlock faults the pages in itself), CSV reports `locked_dense_MiB`.
- **Still needs the device A/B on gpt-oss-120b** — that is what decides whether the default should ever flip. Compare `majflt/tok` and steady-state `compute_ms` against the committed route-trace run.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -335,6 +335,8 @@ static void print_usage(const char * argv0) {
         "      --io-threads N      parallel expert-read lanes [1..%d] (default 4)\n"
         "      --no-odirect        do not bypass the page cache\n"
         "      --no-warm-dense     skip the load-time sweep that page-caches the non-expert weights\n"
+        "      --lock-dense        pin the non-expert weights in RAM (mlock) so expert traffic cannot\n"
+        "                          evict them; costs cache budget — for >RAM models, see the docs\n"
         "      --load-all          debug: read ALL experts each token (A/B baseline)\n"
         "      --force-cache       allow a cache-mb in the pathological band\n"
         "      --overlap           overlap async expert reads with FFN compute (needs the fork)\n"
@@ -403,6 +405,8 @@ int main(int argc, char ** argv) {
             cfg.moe.o_direct = false;
         else if (a == "--no-warm-dense")
             cfg.moe.warm_dense = false;
+        else if (a == "--lock-dense")
+            cfg.moe.lock_dense = true;
         else if (a == "--load-all")
             cfg.moe.load_all = true;
         else if (a == "--force-cache")
@@ -511,9 +515,8 @@ int main(int argc, char ** argv) {
     // occupancy = CPU-time ÷ (wall × threads): ~1 is compute-bound, well under 1 is a throttled or
     // preempted core; major faults/token > 0 means dense weights re-faulted from flash inside decode.
     if (s.cpu_s_per_token > 0.0 || s.majflt_per_token > 0.0) {
-        const double occ = s.s_per_token > 0 && cfg.n_threads > 0
-                               ? s.cpu_s_per_token / (s.s_per_token * cfg.n_threads)
-                               : 0.0;
+        const double occ =
+            s.s_per_token > 0 && cfg.n_threads > 0 ? s.cpu_s_per_token / (s.s_per_token * cfg.n_threads) : 0.0;
         std::printf("compute: %.1f%% CPU occupancy (%.4f cpu-s/token over %d threads), %.2f major faults/token\n",
                     occ * 100.0, s.cpu_s_per_token, cfg.n_threads, s.majflt_per_token);
     }
--- a/core/include/bmoe/config.h
+++ b/core/include/bmoe/config.h
@@ -64,6 +64,30 @@ struct MoeStreamConfig {
     // moves that cost into load_seconds at sequential-read bandwidth. Off for A/B measurements.
     bool warm_dense = true;
 
+    // Pin those same dense regions into RAM (mlock) after warming them, so the kernel cannot reclaim
+    // them again. Warming alone only fills the page cache, which is reclaimable *by definition*: on
+    // a model whose expert traffic dwarfs RAM the streamer's own reads evict the dense weights right
+    // back out and they re-fault from flash inside the next decode — measured at 6.0k→8.4k major
+    // faults per token on gpt-oss-120b, where the wait is billed to "compute" because it happens
+    // under llama_decode (see docs/bench-data/2026-07-15-route-trace/).
+    //
+    // OFF by default: this is a >RAM knob, not a free win. The pinned RAM has to come from somewhere,
+    // and on an auto budget it comes out of the expert cache — pinned pages leave MemAvailable, so
+    // the budget probe below sizes the cache around them. That trade is already measured: reserving
+    // Gemma's ~1.2 GiB of dense cost budget 4000→2909 MiB, hit 82.9→73.5 % and tok/s 4.76→3.76
+    // (docs/bench-data/2026-07-14-warmup/). Worth it only where cache hit rate is cheap to sell —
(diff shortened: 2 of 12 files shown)


Helldez:
## Android cross-compile validated

The POSIX branch of `platform_io.cpp` had never been compiled — the host gates only exercise the Windows `#if`. Closed that gap:

- **`bmoe-cli` cross-compiles clean** for `arm64-v8a` (NDK r26.1, android-29, `armv8.2-a+dotprod+fp16`), no errors or warnings from `platform_io.cpp` / `expert_stream_source.cpp`.
- **The lock path is live on bionic, not compiled out.** `mlock`/`munlock`/`getrlimit`/`setrlimit` all resolve at link time, and a compile-time assert confirms **both `_POSIX_MEMLOCK_RANGE` and `RLIMIT_MEMLOCK` are defined** — worth checking explicitly, because if `_POSIX_MEMLOCK_RANGE` were absent `vm_lock` would have returned 0 and the whole flag would have been a silent no-op on the only platform that matters.

Still open, and only the device can answer: **what `RLIMIT_MEMLOCK`'s hard limit actually is** for an adb-shell process on the OnePlus. `vm_lock_limit_raise()` lifts the soft limit to the hard one, and the log line reports how many MiB it managed vs how many it wanted — if that reads `N of 790 MiB pinned (lockable cap ...)`, the cap is the story and the A/B result says nothing about the idea's merit.

Helldez:
## `mlock` cannot work on Android — measured, don't spend the A/B

This PR's own open question ("what is the real `RLIMIT_MEMLOCK` on the device — if the log says
`N of 790 MiB pinned`, the cap is the story") now has an answer, from `/proc/<engine pid>/limits` on
the OnePlus 15R:

```
Max locked memory         65536                65536                bytes
```

**64 KB, soft and hard.** That is 0.003% of the 1.77 GB dense set. AOSP's `init.rc` sets it
(`setrlimit memlock 65536 65536`), and `vm_lock_limit_raise()`'s `setrlimit` cannot lift the *hard*
limit without `CAP_SYS_RESOURCE`, which an app uid does not have. The chunked `mlock()` loop will pin
one 64 KB chunk and stop; expect `bmoe: dense lock — 0 MiB pinned`.

Corroboration: llama.cpp's `--mlock` fails on Android for exactly this reason, and its advice
("Try increasing RLIMIT_MEMLOCK, `ulimit -l` as root") is unactionable on a phone. ExecuTorch
explicitly maps Android to `NoMlock`. No on-device runtime surveyed pins weights on Android.

So the A/B this PR is waiting for would measure a no-op against noise. Options: drop the mlock path
on Android and keep the host path, or retire the flag.

Worth separating two things this PR's framing conflates: the **problem** it identifies is real and
well-measured (the route trace's 267 → 8431 major faults/token, with `compute_ms` tracking it
539 → 1767 ms — dense weights re-faulting inside `llama_decode`). Only the **remedy** is impossible.
The Gemma 4.76 → 3.76 figure is also not a measurement of this PR — it is the "reserve dense in the
budget" experiment from `docs/bench-data/2026-07-14-warmup/`, a different mechanism.

Full mechanics, with what is measured vs inferred: `docs/android-memory.md` on
`feat/rewarm-after-reclaim` (#28). Short version — nothing an unprivileged Android app can call
protects anonymous memory: memcg is v1 and root-only, `memory.min`/`low` are v2-only, MGLRU is
vendor-disabled, `/proc/<pid>/reclaim` is absent, `oom_score_adj` only affects kill selection. The
one real lever is `MADV_COLD` (5.4+, unprivileged): you cannot stop reclaim, but you can choose its
victims.

Helldez:
Closing this: the premise does not survive measurement on the target device.

`--lock-dense` is built on the assumption that the dense weights can be pinned against reclaim. They cannot. `RLIMIT_MEMLOCK` on the OnePlus 15R is **65536 bytes, hard** — not a soft limit this process can raise, which is why `vm_lock_limit_raise()` cannot help. That ceiling lets the flag pin roughly **0.003% of a 2 GB cache**. The rationale in `config.h:47-57` argues the flag is "a good bet" for gpt-oss-120b specifically; that is the case where the measurement most clearly says otherwise.

The code itself is sound and degrades safely — a refused `mlock` leaves the pages merely warm, so nothing breaks. But shipping a flag that cannot do what its name says, and documenting it as a lever in `adaptive-cache.md`, would put the docs in direct contradiction with #30, which measures the ceiling. A flag whose honest help text would read "pins 0.003% of your cache" is not a flag.

The reason is preserved where it is useful: #30 (`docs/android-memory.md`) records the `RLIMIT_MEMLOCK` finding, why `llama.cpp`'s own `--mlock` fails here for the same reason, and that ExecuTorch maps Android to `NoMlock` rather than fight it. That document outlives this branch.

Not cherry-picking the `dense_file_ranges()` / `warm_dense_regions(fd, ...)` refactor: it exists only to serve this flag, and it also moved the warm sweep ahead of the adaptive budget probe — a behavioural change that was never measured on its own. If a future change needs that split, it should arrive with its own measurement.

What would reopen this: a target where `RLIMIT_MEMLOCK` is actually raisable, or a way to protect the dense weights that does not route through `mlock` at all. The second one is the interesting question, and `android-memory.md` frames it — the expert cache is anonymous memory holding a copy of bytes that are already in the gguf, which is the form the kernel *must* write out to reclaim. That is the cost worth attacking.

Branch `feat/lock-dense` is deleted; the work is recoverable from this PR's history if the premise ever changes.

PR #26, opened 2026-07-15
feat(metrics): --compute-trace and --io-trace decompose the decode

Helldez:
## Why

`compute_ms` is a **residual** (`wall − io − mgmt`), so everything the engine does not itself clock is pooled into it: page faults, scheduler stalls, and the actual matmuls. A residual cannot say which it is.

That is exactly why gpt-oss-120b reads as "compute-bound" at 1.7 s/token while faulting 8.4k pages/token — the flash wait is billed to compute because it happens under `llama_decode`. And the flash floor beneath it is unknown: the aggregate 760 MiB/s sits far below the drive's sequential ceiling, but nothing says whether that is latency, request size, or idle lanes.

## `--compute-trace` — compute measured, not inferred

Returning `true` from the eval callback makes ggml compute exactly up to that node, **synchronize**, then call back — so the wall delta between consecutive boundaries is that node's real compute time. Sampling major faults across the same boundaries attributes the >RAM stall to the node that paid it.

**No llama.cpp patch**: this rides the public `cb_eval` ABI, whose ask/no-ask contract already specifies the isolation ([`ggml-backend.cpp:1682-1713`](third_party/llama.cpp/ggml/src/ggml-backend.cpp)).

The barrier that makes it measurable also inflates it (no operator coalescing), so it is a **diagnostic — a traced run is not a benchmark run**, and only the shares are meaningful. Unlike the other traces it does **not** need `--moe-stream`: it times the graph, so a dense mmap baseline can be traced and compared against a streamed run.

Node classification stays **out of the engine** — which node is attention vs dense FFN vs expert matmul is naming policy that varies by architecture, so rows carry the raw `op`/`name` and the script classifies.

## `--io-trace` — the flash floor

One row per `pread`: latency, requested vs aligned bytes, lane, and the `(layer, expert, projection)` it serves — values already computed at every enqueue site and until now discarded. Reveals alignment waste, the latency distribution, per-lane utilization, and the **adjacency** the roadmap's read-coalescing / expert-contiguous-layout items assume.

## Tooling

`scripts/decode-analyze.py` (stdlib only, mirroring `route-analyze.py`):

```bash
scripts/decode-analyze.py compute ct.csv --layers   # share by op, fault attribution, by layer
scripts/decode-analyze.py io io.csv --adjacent      # latency percentiles, size/bw, lanes, coalescing ceiling
```

## Verification

- Host build clean (0 warnings), **all 4 byte-identity gates pass**.
- Driven end-to-end on both gate models: **generated text, cache hit % and bytes read are identical with the traces on** — they observe only.
- Compute trace verified working **without** streaming (dense baseline, 1890 rows).
- Both analysis views produce correct output on real traces.

Also corrects `docs/telemetry.md`, which claimed measuring compute would require patching the submodule.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -13,6 +13,7 @@
 #include "bmoe/recipe.h"
 #include "bmoe/metrics.h"
 #include "bmoe/route_trace.h"
+#include "bmoe/decode_trace.h"
 
 #include <atomic>
 #include <condition_variable>
@@ -350,6 +351,8 @@ int main(int argc, char ** argv) {
     RunConfig cfg;
     std::string csv_path;
     std::string route_trace_path;
+    std::string compute_trace_path;
+    std::string io_trace_path;
     bool session_mode = false;
 
     for (int i = 1; i < argc; ++i) {
@@ -385,6 +388,10 @@ int main(int argc, char ** argv) {
             csv_path = next("--csv");
         else if (a == "--route-trace")
             route_trace_path = next("--route-trace");
+        else if (a == "--compute-trace")
+            compute_trace_path = next("--compute-trace");
+        else if (a == "--io-trace")
+            io_trace_path = next("--io-trace");
         else if (a == "--moe-stream")
             cfg.moe.enabled = true;
         else if (a == "--cache-mb") {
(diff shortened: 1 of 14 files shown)


Helldez:
Host gates run locally on this branch **rebased onto current main** (CI is still blocked by the Actions billing issue, so this is the only proof available):

```
1/4 moe_make_tiny_model_qwen3moe .....   Passed
2/4 moe_gates_qwen3moe ...............   Passed
3/4 moe_make_tiny_model_gemma4 .......   Passed
4/4 moe_gates_gemma4 .................   Passed

100% tests passed, 0 tests failed out of 4
```

Hard rule 5 applies here because this touches `expert_stream_source` and `router_hook`: `bmoe_moe_gates` proves streamed output is byte-identical to resident, and it holds.

The rebase onto main is a no-op for correctness — main has not touched `core/`, `cli/` or `tests/` since this branch was cut (its four intervening commits are docs, scripts and Android only), so the gate result carries over unchanged.

PR #28, opened 2026-07-15
Rewarm after reclaim — a measured failure, kept as a baseline

Helldez:
> **Draft: parked experiment, not a merge candidate.** The feature ships **off by default** because it was measured on device and it does not work. It is kept open so the primitives, the gate and the telemetry stay available as the baseline a real fix must beat.

## What it does

`--rewarm` / `--rewarm-threshold-mb N`: before a generation, if `/proc/self/status:VmSwap` says the kernel reclaimed ≥ threshold MiB while the session idled, bulk-restore it — `MADV_WILLNEED` over the expert buffers in address order, plus a dense re-sweep — instead of paying it back one major fault at a time inside the decode. Its seconds are reported as `rewarm_s` and folded into TTFT, not the decode.

## Why it is off

It restores 2.0 GiB in **6.4 s**, and the kernel takes it back **8 seconds later, mid-decode**, with the turn still at **0.3 tok/s**. That 8 s is not a bug in the code — it is a property of the LRU. `MADV_WILLNEED` on anon swaps pages back via `read_swap_cache_async`, which puts them on the **inactive** list; nothing marks them accessed. So a bulk restore hands the kernel a pile of pages at the *front* of the eviction queue, and at swappiness 160 with `inactive_ratio = 1` the next pass comes almost immediately.

The mechanics are now documented on `main` in [`docs/android-memory.md`](docs/android-memory.md) (landed via #30), which is where the reusable part of this work ended up.

The corollary is the useful part, and it points at the next attempt: **a restore is only worth anything if the pages get hit right after** — which makes its value a function of the hit rate, and predicts it helps exactly the high-hit-rate models it was not designed for.

## What changed in this revision

Rebased onto current `main`, which absorbed four PRs in the meantime. The branch is now **4 commits, all rewarm** — down from 5 files' worth of unrelated work:

- The `docs/android-memory.md` write-up was extracted to #30 (merged). It was the most reusable content here and had no business being hostage to a feature that ships disabled.
- The 500/1000 MiB cache rungs were extracted to #31 (merged) — an independent Settings change.
- Added a commit re-linking `android-memory.md` to the artifacts that live here (`rewarm.md`, the bench-data capture, `scripts/rewarm-ab.sh`), since it had to stand alone on `main`.

The `--lock-dense` branch (#25) was retired, which incidentally removed a **silent build break**: #25 changed `warm_dense_regions`'s signature while this branch's `rewarm()` called the old one, and the two edits auto-merged cleanly into code that would not compile. Verified absent — the declaration and the call site now agree.

## Verification

Built and gated locally against this rebase (CI is still blocked by the Actions billing issue):

```
1/4 moe_make_tiny_model_qwen3moe .....   Passed
2/4 moe_gates_qwen3moe ...............   Passed
3/4 moe_make_tiny_model_gemma4 .......   Passed
4/4 moe_gates_gemma4 .................   Passed

100% tests passed, 0 tests failed out of 4
```

This is the only branch of the recent set that satisfies hard rule 5 on its own terms: it **adds** gate **S4**, which forces the pass with `rewarm_threshold_mb = 0` plus prefetch and asserts streamed == resident across two generates (`S4a`, `S4b`). The conflict with #26's `set_io_trace`/`take_io_trace_rows` in `expert_stream_source.h` was textual — both add public methods at the same anchor — and both are kept.

## What would close this

Either a trigger that also sees file-backed reclaim (`RssFile` from the same `/proc/self/status` read is the obvious candidate — `VmSwap` is blind to the dense case), or an approach that does not fight the LRU at all. `android-memory.md` frames the real question: the expert cache is anonymous memory holding a copy of bytes that already exist in the gguf, which is the one form the kernel must write out to reclaim. That is the cost worth attacking.


Helldez:
Closing this measured failure. Rewarm-after-reclaim was tested and refuted: the kernel reclaims the restored pages mid-decode, so restoring memory can't win. On gpt-oss the 2 GiB expert cache buys only an 8-9% hit rate and pays for it with a memory war. This line of work is superseded by the pressure-aware cache governor (#35), which sizes the cache to what the device actually concedes rather than fighting reclaim. The code stays reachable through this PR as a baseline; no reason to keep it as an open draft.

PR #35, opened 2026-07-15
Pressure-aware cache sizing: find the budget the device concedes

Helldez:
Off by default (`--cache-dynamic`, plus an app switch). Opt-in until the device says otherwise.

## Why

The rewarm result (#28) established that on a >RAM model the engine asks for more memory than the
device concedes, and that restoring the stolen pages cannot win: the kernel takes them back
mid-decode and the turn stays at 0.3 tok/s. The conclusion was that **the ask has to shrink**.

The obvious follow-up — pick a smaller number — is the thing this PR is trying not to do. A cache
budget cannot be a constant: it depends on the model (one gpt-oss token at top-2 routes ~536 MiB,
one Qwen3-30B token at top-8 routes ~1051 MiB), the top-k, the phone, and whatever else the user has
open right now. `--cache-mb 1000` measures well on gpt-oss on this device today, and that is exactly
the kind of fact that does not survive a new model or a new phone.

## What it does

Treat the configured budget as a **ceiling**, and find the largest cache the device actually
concedes.

Not by asking the OS — on Android every signal for this is unusable: `MemAvailable` counts our own
mmap'd weights as free (which is why `--cache-mb auto` over-asks by construction), PSI is
SELinux-blocked for apps, `onTrimMemory` is late and never signals relief, `mlock` is capped at
64 KiB. So the engine watches **its own pages** instead — the one thing every device lets a process
ask about, no permission and no vendor cooperation required:

- **Sense** — `mincore()` over the cache's own buffers, walking the LRU cold end (the pages reclaim
  takes first). Absolute signal: our pages are in RAM or they are not. Plus `getrusage` major-fault
  deltas as the reflex between samples (already measured for telemetry).
- **Decide** — AIMD with hysteresis (`bmoe/cache_governor.h`). Pure policy: no syscalls, no clocks,
  no llama.cpp. The shape TCP uses against an unobservable bottleneck.
- **Act** — the existing `set_cache_budget`, ticked once per token *after* the decode: the only
  point where nothing is staged for a generation in flight.

## The floor is measured, not guessed

`cache_min_mb` (1500) encodes a real bound — below one token's routed working set, every token
evicts what the next one needs — but encodes it as a constant that is model-specific, which is why
the app's 500/1000 rungs need `--force-cache` to measure the very thing the floor is guessing at.

A token's pass over the stack is monotonic in layer index, so the bytes demanded between two visits
to the same layer **is** the working set. The streamer now measures it (`token_demand_MiB`), and it
is what floors the governor. That is what makes this work on a model nobody has benchmarked yet.

## Verification

- New unit tests (`tests/cache_governor_test.cpp`): the governor is pure policy, so the device is a
  function and the properties that matter are testable without a phone — converges just under a
  hidden concession **and stops**, never cuts below a token's demand, a missing sample is not read as
  pressure. They caught a real bug: the cooldown, starting unspent, delayed the *first* cut, leaving
  a session opened onto an already-pressured device inert exactly when it mattered.
- Byte-identity gates green. Output verified identical across static / dynamic / dynamic+overlap /
  cache-off on both architectures — the loop changes what is resident, never what is computed.
- Host build: no residency signal, so the loop stays calm and the budget behaves exactly as a static
  one. That is the intended degradation, not an accident.
- One drive-by: the `occ` line in `cli/main.cpp` already violated clang-format on main; the format
  gate runs whole files, so it is fixed here.

## Not verified yet

**On-device A/B is outstanding** — that is the whole point of the switch. The claim to test on
gpt-oss: the governor should match the best static budget (1000 MiB today) *without being told it*,
and not regress Qwen/Gemma, where the cache genuinely pays (53-82% hit) and `auto` currently
over-asks. It stays off by default until those numbers exist.

Also carried but unwired: `MADV_PAGEOUT` was considered for the evicted tail and left out —
`MADV_DONTNEED` already frees those pages more cheaply. It would only make sense against the
*retained* cold tail, which is a separate experiment.

See [docs/pressure.md](docs/pressure.md).

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -260,10 +260,11 @@ static int run_session_loop(const RunConfig & cfg, IMetricsSink * sink, IRouteTr
                 std::printf("BMOE_LOAD {\"mb\":%.2f,\"ms\":%.1f}\n", m.read_bytes / (1024.0 * 1024.0), m.io_ms);
             std::printf("BMOE_PROGRESS {\"step\":%d,\"steps\":%d,\"wall_ms\":%.1f,\"io_ms\":%.1f,"
                         "\"compute_ms\":%.1f,\"mgmt_ms\":%.1f,\"stall_ms\":%.1f,\"read_mb\":%.2f,"
-                        "\"cache_hit_pct\":%.1f,\"majflt\":%llu,\"cpu_ms\":%.1f,\"text\":\"%s\"}\n",
+                        "\"cache_hit_pct\":%.1f,\"majflt\":%llu,\"cpu_ms\":%.1f,\"resident_frac\":%.3f,"
+                        "\"text\":\"%s\"}\n",
                         m.step, m.steps, m.wall_ms, m.io_ms, m.compute_ms, m.mgmt_ms, m.stall_ms,
                         m.read_bytes / (1024.0 * 1024.0), m.cache_hit_pct, (unsigned long long) m.majflt, m.cpu_ms,
-                        json_escape(m.text).c_str());
+                        m.resident_frac, json_escape(m.text).c_str());
             std::fflush(stdout);
         };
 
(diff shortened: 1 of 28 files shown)


Helldez:

PR #38, opened 2026-07-16
gov2 cache governor — attribution, plateau, and a real off state

Helldez:
## What

The second-generation cache governor, default off behind `--cache-gov2`. Split out of the previously
bundled commit so it has its own history and its own on-device A/B outcome, independent of
`--dense-odirect` (which stacks on this branch).

On top of `--cache-dynamic`'s shrink-to-what-the-device-concedes loop it adds:

- **Attribution** of a token's fault load to its culprit (our cache, the dense weights, or the
  anon/zram war), so a cut fires only when the cache is actually the cause.
- A **hit-rate plateau** that replaces a hand-set ceiling — growth stops where more budget stops
  buying hits (two binomial standard errors).
- A real **OFF state**: on a >RAM model that routes more than the device will hold, the LRU earns no
  hits and its commit/evict churn feeds the war, so the governor demotes the cache to shared-slot
  mode and re-arms only when the device frees the room.
- An **entry test** that clamps the starting budget before prefill can knock the device over
  (`MemAvailable` counts our own mmap'd weights as free, so it is discounted by the dense bytes).

## Mechanism

`ExpertStreamSource` gains a runtime LRU ⇄ shared-slot switch (`set_cache_mode`), with the slot
buffers allocated lazily on the first demotion; the slot-mode predicate becomes a mutable
`slot_mode_` member so the mode can flip between tokens. `Session` feeds the governor per-token
signals (dense residency, swap delta, hit rate, token demand) and applies its decisions once per
token, where no decode is in flight. `stats()` exposes `total_expert_bytes` (the physical cache
ceiling). No llama.cpp patch; no env vars in the library; no model constants.

## Telemetry

`gov_state` / `gov_war` columns and a `cache_gov2` preamble flag in the CSV.

## Tests

- **S4** (byte-identity): the runtime mode switch (LRU → slots → LRU) changes not a single output
  byte, serial and overlap, on both tiny fixtures.
- `cache_governor_test.cpp`: unit-tests the policy (demote/promote/plateau/attribution) against
  synthetic device signals.

Host build + full `ctest` green (G1–G5, S1–S4, `cache_governor`). Default off pending the on-device
A/B on both phones. See `docs/pressure.md`.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -338,6 +338,9 @@ static void print_usage(const char * argv0) {
         "      --cache-dynamic     treat --cache-mb as a ceiling and find the budget the device\n"
         "                          concedes: shrink when reclaim starts taking the cache, grow back\n"
         "                          when it stops (needs the cache; see docs/pressure.md)\n"
+        "      --cache-gov2        --cache-dynamic plus a hit-rate plateau, fault attribution, and an\n"
+        "                          OFF state (demote the cache to slots on a >RAM model where cutting\n"
+        "                          cannot end the war); default off, promoted per model+device by A/B\n"
         "      --io-threads N      parallel expert-read lanes [1..%d] (default 4)\n"
         "      --no-odirect        do not bypass the page cache\n"
         "      --no-warm-dense     skip the load-time sweep that page-caches the non-expert weights\n"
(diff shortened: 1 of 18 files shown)


Helldez:

PR #39, opened 2026-07-16
--dense-odirect — stream dense weights via O_DIRECT into anon buffers

Helldez:
## What

`--dense-odirect` (default off, in-app toggle "Dense via O_DIRECT"): read the dense (non-expert)
weights once via O_DIRECT into aligned anonymous buffers and rebind their tensors onto them, instead
of leaving them mmap'd file-backed. A reclaim of anonymous memory swaps to zram (fast decompress)
rather than dropping to a scattered 4 KiB flash refault. A strict A/B lever — the analysis says its
upside is narrow (Q4 is near-incompressible; it is a wash unless the model is actively losing its
dense set to reclaim).

**Stacked on `feat/cache-gov2`** — base this PR on that one; it was split out of the same original
commit so the two features have independent A/B outcomes.

## Mechanism (no llama.cpp patch)

Reuses the expert-capture seam: `RouterHook` records every non-expert weight leaf in the warm-up
graph (`captured_weights()`); `Session` pairs those with their gguf offsets, minus the streamed
expert set, and hands them to `ExpertStreamSource::read_dense_odirect()`, which reads each tensor
chunked on lane 0 before the workers start and rebinds `->data`. Buffers freed in `shutdown()`.

## Hardening for a trustworthy A/B (second commit)

- **Drop the doubled residency (P4).** The capture warm-up faults the dense pages in mmap-resident
  before they are copied into the anon buffers, so for a moment the set is resident twice — right
  before prefill, when the gov2 entry test reads `MemAvailable`. After the rebind, the now-
  unreferenced mmap pages are dropped with `MADV_DONTNEED` via a new file-mapping-safe primitive
  `pio::vm_drop_file_pages` (POSIX `madvise`, Windows no-op), page-aligned inward off
  `/proc/self/maps` so a page shared with a still-resident tensor is never dropped.
- **Keep the sensor live (P5).** `sample_dense_residency` samples the anon buffers directly instead
  of going inert, so `dense_resident_frac` keeps its meaning (whether zram is holding the dense set)
  under the flag rather than reading unmeasured.

## Tests

- **G6** (byte-identity): the rebind changes not a single output byte vs the mmap-resident reference.
- **G7** (byte-identity): the same holds with O_DIRECT off (the dense read then lands buffered).

Host build + full `ctest` green (G6, G7, S1–S4 on both fixtures). Default off pending the on-device
A/B on both phones. See `docs/pressure.md`.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -341,6 +341,7 @@ static void print_usage(const char * argv0) {
         "      --io-threads N      parallel expert-read lanes [1..%d] (default 4)\n"
         "      --no-odirect        do not bypass the page cache\n"
         "      --no-warm-dense     skip the load-time sweep that page-caches the non-expert weights\n"
+        "      --dense-odirect     read the dense weights via O_DIRECT into our buffers (experiment)\n"
         "      --load-all          debug: read ALL experts each token (A/B baseline)\n"
         "      --force-cache       allow a cache-mb in the pathological band\n"
         "      --overlap           overlap async expert reads with FFN compute (needs the fork)\n"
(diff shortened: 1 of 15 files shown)


Helldez:

PR #40, opened 2026-07-16
Harmonize dense/expert O_DIRECT: FileReader + DenseWeights modules

Helldez:
## What

Harmonizes how the dense and expert weights do O_DIRECT I/O, and untangles the dense-weights
handling from the expert streamer — the modularity + "which O_DIRECT, dense or expert or a mix"
question that motivated this. Pure structural refactor: **byte-identity gates G1–G7 + S1–S4 green on
both fixtures at every step.**

Stacked on `feat/dense-odirect` (#39).

## The seam, before → after

Before, `ExpertStreamSource` owned everything: the fd pool + bounces + O_DIRECT verify, `read_slice`,
the dense warm sweep, the `--dense-odirect` anon read+rebind+drop, and the dense residency sensor —
and every read rode one `o_direct_` flag, so `--no-odirect` silently changed the dense read too, and
"which bytes are dense" was computed three separate times.

After, three layers:

- **`FileReader`** (`core/src/io/file_reader.{h,cpp}`) — a pooled positioned reader: fd pool, bounces,
  aligned-window mechanics, and the O_DIRECT open+verify+fallback. **O_DIRECT is a property of the
  reader**, chosen by whoever opens it. Knows nothing about experts or dense.
- **`DenseWeights`** (`core/src/moe/dense_weights.{h,cpp}`) — the non-expert weight policy, its
  buffers, and the residency sensor, with one `byte_ranges()` definition shared by all consumers. It
  opens its **own** `FileReader`, so the dense O_DIRECT choice is independent of the expert stream's —
  the experts can stream buffered while the dense set is pulled cache-bypassing, or the reverse.
- **`ExpertStreamSource`** — back to expert streaming + the LRU/slot cache; it *composes* a reader and
  a `DenseWeights`. ~250 lines of non-expert concern left it.

## Policy consolidation

The two bools `warm_dense` + `dense_odirect` collapse into one enum:

```
DenseWeightsMode { Mmap, Warmed (default), Anonymous }
```

CLI gains `--dense-weights mmap|warm|anon`; `--no-warm-dense` (→`Mmap`) and `--dense-odirect`
(→`Anonymous`) stay as **deprecated aliases**, so scripts and the app keep working. The CSV still
emits `warm_dense`/`dense_odirect`, derived from the resolved policy — telemetry and the Android
reader unchanged.

## Answering the original question

Priority isn't dense-O_DIRECT *or* expert-O_DIRECT — it's the **mix**, and this is what makes it
expressible: experts are always direct-streamed (the hot path), and the dense set picks its own policy
per model, with the two O_DIRECT decisions genuinely independent. `G7` proves it: `dense=anon` with
the expert stream's O_DIRECT *off* still produces byte-identical output.

## Not in scope (a separate follow-up)

The measurements say the adaptive cache / governor doesn't pay for the >RAM case this engine exists
for (cache-off is the ceiling). Removing that machinery is a separate cut, deliberately kept out of
this refactor — this branch stays governor-agnostic and would survive it unchanged.

Host build + full `ctest` green. See `docs/pressure.md`, `docs/architecture.md`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -11,8 +11,11 @@ we do not fork llama.cpp. See `docs/architecture.md` and `docs/seam.md`.
 ## Project map
 
 - `core/include/bmoe/` — ports (interfaces) + config. Pure policy, no llama.cpp include.
-- `core/src/io/` — `platform_io`: cross-platform O_DIRECT reads + reserve/commit/evict VM.
-- `core/src/moe/` — `gguf_offsets`, `arch_registry`, `expert_stream_source`, `router_hook`.
+- `core/src/io/` — `platform_io` (cross-platform O_DIRECT reads + reserve/commit/evict VM);
+  `file_reader` (pooled positioned reader, per-consumer O_DIRECT — used by both the expert stream
+  and the dense loader).
+- `core/src/moe/` — `gguf_offsets`, `arch_registry`, `expert_stream_source`, `router_hook`;
+  `dense_weights` (the non-expert weight policy: mmap / warm / anon, plus the residency sensor).
 - `core/src/engine/runtime.cpp` — composition + greedy generation loop.
 - `cli/main.cpp` — `bmoe-cli`; the ONLY place environment variables are read.
 - `third_party/llama.cpp` — stock upstream submodule.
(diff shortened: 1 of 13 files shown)


Helldez:

PR #41, opened 2026-07-16
Retire the adaptive cache governor; keep the fixed LRU + auto sizing

Helldez:
## What

Retires the adaptive cache **governor** — the on-device measurements settled that for the >RAM models
this engine exists for, **cache-off is the ceiling**, and the runtime loop was a net loss. Keeps the
parts that pay: the fixed-budget LRU, one-shot `--cache-mb auto`, cache-off (the default), and the
O_DIRECT dense/expert streaming.

Stacked on `feat/harmonize-io` (#40).

## Why

Measured (gpt-oss at top-4): governed run 0.156 tok/s vs cache-off 0.517. The governor assumed asking
too much costs a war while asking too little costs a few hit-points — but on a >RAM model there's no
budget large enough to earn hits *and* small enough to avoid the reclaim war, so the loop just paid
the sensing cost and churn while converging on what cache-off gives for free.

## Removed

- `cache_governor.{h,cpp}` + `cache_governor_test.cpp` (the decision module).
- `--cache-dynamic` / `--no-cache-dynamic` and the config field; the app's "Pressure-aware cache
  sizing" toggle.
- The runtime sense/resize loop in the streamer: `sample_residency`, `adapt_cache_budget`,
  `probe_tick_`, `resident_frac_`, the per-token governor tick in `Session`.
- CSV `cache_dynamic` column and `RunInfo.cache_dynamic`; `cache_cuts` is now always 0.

## Kept (measured to pay, or cheap + safe)

- **`--cache-mb N`** — the fixed-budget LRU that keeps the hottest experts resident (53–82% hit on
  RAM-fitting Qwen/Gemma). Byte-identity still gated by **G2**.
- **`--cache-mb auto`** — now a **one-shot** sizing from device RAM at load, then fixed (the runtime
  nudge went with `adapt_cache_budget`). `--cache-ceil-mb` / `--cache-floor-mb` still bound it.
- **cache-off (shared-slot)** as the default and the recommendation for >RAM.
- `set_cache_budget()` for an app's `onTrimMemory`; the DenseWeights residency sensor as telemetry.

## App impact (verified)

The "Pressure-aware cache sizing" switch and its `cacheDynamic` setting are removed from `AppSettings`,
`SettingsScreen`, and the metrics screen's config table; the app no longer emits `--cache-dynamic`.
The cache-size picker (fixed / auto) and `--dense-weights` toggle are unchanged. The APK needs a
rebuild to pick this up.

Host build + full `ctest` green (G1–G7, S1–S3 on both fixtures; the governor unit test is gone).
Docs reframed — `pressure.md` keeps the war analysis as the rationale for cache-off. The harmonization
(FileReader / DenseWeights) is untouched: it was built governor-agnostic for exactly this.


Helldez:

PR #83, opened 2026-07-19
fix: honour the Thinking toggle on models whose template ignores enable_thinking

### LINKED ISSUE #82: Thinking toggle is a silent no-op on models whose chat template ignores enable_thinking (LFM2.5)
## Summary

Turning Thinking off in the app (`--no-think` on the CLI) has no effect on LFM2.5-8B-A1B: the
model reasons anyway, and nothing tells the user the setting was ignored.

## Root cause

`--no-think` does exactly one thing — `inputs.enable_thinking = false`
(`core/src/engine/session.cpp:635`) — which llama.cpp forwards to the model's own Jinja chat
template as a template variable (`common/chat.cpp:895`). Whether that suppresses reasoning is
entirely up to the template.

LFM2.5's `chat_template.jinja` never references `enable_thinking`. Its only related variable is
`preserve_thinking`, which controls whether *past* turns' reasoning is re-rendered, not whether the
model reasons on this turn. So the rendered prompt is byte-identical with thinking on and off, and
the model — a reasoning model — reasons.

## Scope: three regimes, not two

| Family | Template honors `enable_thinking` | Outcome |
|---|---|---|
| Qwen3 / Qwen3.6 | yes | toggle works |
| gpt-oss (harmony) | no | works anyway — `session.cpp:651` special-cases it by priming the final channel |
| LFM2 / LFM2.5 | no | **toggle is a silent no-op** |

Gemma 4 does not raise the question (not a reasoning model). The failure is not specific to LFM2.5:
any architecture whose template ignores the variable lands here, and always silently.

## Note on the cost

The reasoning is still parsed out into the collapsible thinking block — the parser gate keys on the
presence of `<think>` in the template source, not on `enable_thinking` (`common/chat.cpp:1708`). So
the answer is not polluted. The real cost is that the model spends its token budget reasoning when
the user asked it not to.

## Trap for whoever fixes this

llama.cpp appears to already offer the detector: `common_chat_templates_support_enable_thinking()`
(`common/chat.cpp:355`). **It does not answer this question.** It returns `params.supports_thinking`,
which the LFM2 handler hardcodes to `true` (`common/chat.cpp:1696`) — it reports "this model can
reason", not "this template honors the flag".

## Direction (not yet decided)

1. Detect generically: render the prompt with `enable_thinking` true and false and compare. Identical
   renders mean the flag has no purchase on this model — surface that in the UI instead of offering a
   toggle that does nothing.
2. Prefill, as already done for harmony: close the reasoning span in the generation prompt
   (`<think></think>`) so the model proceeds to the answer. Must key on the template, not the model
   name (see CLAUDE.md: no model-specific constants).

Prior art in llama.cpp's own server and in other on-device apps is being surveyed before choosing.

## Repro

LFM2.5-8B-A1B-Q5_K_M.gguf, app Thinking toggle off, any prompt. Observed on 0.11.1-dev.

Helldez:
Closes #82.

The Thinking toggle only ever set the chat template's `enable_thinking` kwarg. That is a
*request*, and a template is free to ignore it — LFM2.5's never reads the variable, so it
rendered an identical prompt either way and the model reasoned on, with nothing to tell the
user the request had been dropped.

Following the direction settled in the issue: detection and enforcement are separate
questions, and both are needed. The sampler alone leaves a UI unable to say when the toggle is
real; the probe alone cannot stop a model that has no off switch.

## Detection — `probe_think_control`

Renders the template with thinking on and off and compares. This is the only honest way to ask
the question, and it is what llama.cpp's own template analyzer does internally. The published
`common_chat_templates_support_enable_thinking()` cannot answer it: for any template matching a
known handler `supports_thinking` is a hardcoded literal (`true` at nine sites), reporting
"this model can reason" rather than "this template reads the flag".

Runs once per session, classifying into four regimes:

| `think_ctl` | meaning |
|---|---|
| `template` | renders differently with thinking off — honoured in the prompt (Qwen3/3.6) |
| `forced_final` | harmony/gpt-oss: the engine primes the final channel instead |
| `sampler` | flag ignored, but the model's thinking tags let the engine force the close (LFM2.5) |
| `none` | flag ignored and no tags: the request cannot be honoured |

Reported in `BMOE_READY` so a client can stop offering a control that does nothing. Failures
fail open to `template` — an unknown template is never accused of ignoring the flag.

## Enforcement — `make_think_budget_sampler`

llama.cpp's reasoning-budget sampler, driven by the tags the loaded model reports. It acts on
logits, so it never asks the template's permission — precisely the failure in this issue. With
budget 0 it closes the block on the first token the model opens it with; the model emits
`<think></think>` and the reasoning parser yields an empty block and a clean answer.

It arms only where the probe found no other mechanism, so templates that already honour the
flag and harmony models are untouched. `--reasoning-budget N` generalises it to a cap rather
than all-or-nothing, and rides generate requests as `reasoning_budget`.

Nothing here keys on a model name or a template pattern — the tags, the generation prompt and
the verdict all come from the loaded model (hard rule 4). `common/reasoning-budget.h` is
already in the pinned submodule's build and `llama-common` is already linked, so this is not a
llama.cpp patch and needs no build changes (hard rule 1).

## No regressions

The byte-identity gates pass unchanged, and by construction rather than by luck: they render no
template, so nothing is ever armed and the decode line reduces to exactly the ternary it
replaced. Outside the few forcing steps the sampler is a passthrough and is skipped entirely
rather than applied and trusted to do nothing. One documented deviation from upstream's
composition: forced steps bypass the sampling chain (result-equivalent, since only one
candidate survives), so the dist RNG does not advance on a step whose outcome was never in
question.

## Tests

`tests/think_control_test.cpp` — model-free, mirroring `chat_parse_test`: classifies the three
real vendored templates one per regime (Qwen3 → `template`, LFM2.5 → `sampler` with identical
renders and `<think>`/`</think>` tags, gpt-oss → `forced_final`), plus the no-support case,
the harmony-marker edge cases, fail-open, and the wire-name contract. Local run: all 7 ctest
tests green, including both `moe_gates` fixtures.

## Still owed before release

The on-device A/B through the app: LFM2.5 with the toggle off should produce no thinking block,
toggle on unchanged, and Qwen3.6/gpt-oss unchanged in both states. Host gates prove the wiring,
not the behaviour on a real gguf.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,14 @@ Semantic Versioning.
   the tokens spent inside one reasoning block; `-1` unlimited (default), `0` closes it immediately.
   Enforced on logits, so it applies whatever the template supports.
 
+- **Whitespace before a reasoning block no longer leaves the markers in the answer.** The generated
+  PEG parser anchors its reasoning rule immediately after the generation prompt and makes it
+  optional, so a single leading space or newline stops it matching — and the parse still succeeds,
+  with both markers falling through into the content. Suppressing the block made this visible rather
+  than causing it (an empty block is all markers, with no reasoning text to disguise the leak), so
+  the trim applies to normal reasoning spans too. It fires only when whitespace is all that stands
+  between the stream and the opening tag; an answer that legitimately opens with a blank line keeps it.
+
 ### Notes
 - The byte-identity gates pass unchanged: nothing is armed on the raw-prompt path the gates run, so
   the decode loop reduces to exactly the line it replaced.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,33 @@ Semantic Versioning.
 
 ## [Unreleased]
 
+## [0.13.0] - 2026-07-19
+
+### Fixed
+- **The Thinking toggle is no longer a silent no-op on models whose chat template ignores it**
+  (issue #82). `--no-think` set the template's `enable_thinking` kwarg and stopped there, which is
+  only a *request*: LFM2.5's template never reads the variable, so it rendered the same prompt
+  either way and the model reasoned regardless, with nothing to tell the user. The request is now
+  enforced where the template cannot veto it — a reasoning-budget sampler closes the block on the
+  first token the model opens it with, driven by the thinking tags the loaded model itself reports.
+  gpt-oss keeps its existing primed-final-channel path (it exposes no tags), and templates that do
+  honour the kwarg are untouched, so nothing that already worked changes.
+
+### Added
+- **`think_ctl` in `BMOE_READY`**: which mechanism will honour a `think:false` request for the
+  loaded model — `template`, `forced_final`, `sampler`, or `none`. Decided by rendering the chat
+  template with thinking on and off and comparing, because llama.cpp's published `supports_thinking`
+  cannot answer it (per-handler it is a hardcoded literal meaning "this model can reason"). Probe
+  failures report `template`, so an unknown template is never wrongly accused. The Android app uses
+  it to describe the Thinking switch honestly instead of offering a control that does nothing.
+- **`--reasoning-budget N`** (and the matching `reasoning_budget` field on generate requests): cap
+  the tokens spent inside one reasoning block; `-1` unlimited (default), `0` closes it immediately.
+  Enforced on logits, so it applies whatever the template supports.
+
+### Notes
+- The byte-identity gates pass unchanged: nothing is armed on the raw-prompt path the gates run, so
+  the decode loop reduces to exactly the line it replaced.
+
 ## [0.12.0] - 2026-07-19
 
 ### Added
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -188,6 +188,7 @@ struct SessionCmd {
     int n_predict = 128;
     bool think = true;
     bool clear_kv = true;
+    int reasoning_budget = -1;
 };
 
 // Interactive session: keep the model loaded and the expert cache warm across prompts, reading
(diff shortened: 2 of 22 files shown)


Helldez:
Closing this. On-device testing showed the approach does not hold up, and the patches I added to
rescue it made the branch worse rather than better.

**What the measurement showed.** With `--reasoning-budget 0` and an empty forced message â€” the
upstream default â€” LFM2.5-8B-A1B does not accept the forced close. It re-opens the block, gets
closed again, and after the second denial abandons the tags and reasons in plain prose, which lands
straight in the answer along with a stray closing tag. That is worse than not suppressing at all:
before this branch the reasoning was at least tagged, and so collapsible in the UI.

**Where I went wrong after that.** Rather than treating that as a verdict on the approach, I patched
around the symptoms:

- A hardcoded English sentence injected into the forced close, chosen by trial and error against one
  model. Upstream deliberately exposes this as a flag with an empty default instead of picking the
  wording for you. Mine was invented, monolingual, and buried in the engine â€” squarely against the
  no-hardcoding rule in CLAUDE.md.
- String surgery (`skip_forced_reasoning`) to cut leading reasoning blocks out of the model's output,
  because the parser strips only the first one. Trimming model output by hand is the kind of thing
  that rots on the next submodule bump.

**Also worth re-examining:** whether the render-twice probe earns its place. llama-server runs no
probe at all â€” it checks `thinking_end_tag` for emptiness and wires the sampler. A four-state
classifier may be answering with a mechanism what upstream answers with a conditional.

What survives is the diagnosis and the measurements, which are recorded in #82. The branch stays for
reference; it should not be merged. Reopening #82 to be approached differently.

PR #85, opened 2026-07-19
fix(chat): honour "thinking off" on templates that ignore enable_thinking

### LINKED ISSUE #82: Thinking toggle is a silent no-op on models whose chat template ignores enable_thinking (LFM2.5)

Helldez:
Closes #82 (pending the on-device check below).

`enable_thinking` is a *request* to the model's chat template, and a template is free to ignore it.
LFM2.5's never reads it, so the rendered prompt was byte-identical with Thinking on and off, the
model reasoned anyway, and nothing said the setting had been dropped.

## What changes

**Detection is measured, not declared.** At `open()` the template is rendered with the flag on and
off and the two prompts compared, then rendered once more with a continuation applied. That answers
the only question that matters — *does this template react?* — for any model, in any language, with
no model names anywhere. `common_chat_templates_support_enable_thinking()` cannot answer it: per
handler it is a hardcoded literal reporting "this model can reason", not "this template reads the
variable".

**Enforcement is llama.cpp's continuation hook.** A synthetic trailing assistant message plus
`continue_final_message` makes upstream's own per-template handler render *that family's* "reasoning
is over" span into the prompt, so the model resumes at the first token of its answer with the
reasoning already behind it. This is exactly what a template that implements the toggle natively
does — Qwen3 renders `<think></think>` for `enable_thinking=false` — and it is the technique the
DeepSeek-R1 community settled on for models whose template lacks the flag. It needs no cooperation
from the template and **no sampler**, so it works on the greedy path the byte-identity gates run on.

The engine writes about 90 lines and spells out no markers of its own.

## The harmony hardcoding is gone

Priming gpt-oss to answer without reasoning used to be a literal `<|start|>assistant` suffix test
plus a literal `<|channel|>final<|message|>` appended in the decode path — the last model-specific
constant in `generate()`. gpt-oss now takes the same generic path as every other family and resumes
at the same point in the prompt (upstream's handler renders an empty analysis turn ahead of it,
which is if anything more harmony-correct). A submodule bump that changes those markers no longer
needs an engine change.

## What is deliberately NOT here

Forcing the reasoning block closed on logits. Measured on-device in the closed #83, it made LFM2.5
strictly worse: the model reopened the block, and once denied twice it abandoned the tags and
reasoned in plain prose straight into the answer — 166 tokens of reasoning moved *out* of a
collapsible block and *into* the reply. Suppression has to land in the prompt, before the model
commits to reasoning, not mid-generation. No output string surgery either, and no invented English
sentence in the engine.

## When the model refuses

Reported, not fought. `BMOE_READY` gains `think_ctl` (`template` | `prefill` | `none`), documented
in `docs/telemetry.md`; the app shows the Thinking switch **disabled with the reason** when a model
reasons no matter what it is asked, instead of offering a control that does nothing.

## Tests

`tests/think_control_test.cpp` — no model needed, runs unconditionally, drives real templates from
the vendored submodule:

| case | asserts |
|---|---|
| Qwen3 | probed as `template` (flag honoured, engine adds nothing) |
| LFM2.5 | flag is inert; probed as `prefill`; **span asserted CLOSED and not reopened** |
| gpt-oss | probed as `prefill`; prefill moves past the reasoning channel |
| plain template | probed as `none` |
| probe fails | falls open to `template`, i.e. pre-existing behaviour |

Expected shapes are derived from each template's own thinking-on render, so a submodule bump that
changes a family's markers updates the expectations with it.

Locally: 7/7 ctest green (byte-identity MoE gates included), clang-format 18 clean, arm64
cross-build clean, app Kotlin compiles. CI still cannot run (Actions billing).

## Still owed before merge — on-device, via the app

Host gates prove the prompt is shaped right; they cannot prove how a model *reacts* to it. That is
exactly the gap that sank #83, so:

- **LFM2.5, Thinking off** — the answer must contain no prose reasoning and no stray markers. This
  is the run that decides the PR. If the model reasons anyway despite the closed span, the honest
  outcome is to classify it `none` and let the UI say so — not to fight it.
- **Qwen3.6** — unchanged behaviour (`template` path, untouched).
- **gpt-oss** — same answer quality as today, since this replaces a currently-working path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,14 @@ Semantic Versioning.
   variable `enable_thinking` and stopped there — but that variable is only a *request* to the
   model's chat template, and many templates never read it. LFM2.5's is one: the rendered prompt came
   out byte-identical either way, the model reasoned on, and nothing reported that the setting had
-  been dropped. The engine now renders the template at load to find out which case a model is in,
-  and where the flag is inert it closes the reasoning span in the prompt instead, using llama.cpp's
-  own continuation hook so the markers for each family come from upstream's per-template handler
-  rather than from this engine. Models that support neither are reported as such
-  (`think_ctl` on `BMOE_READY`, see `docs/telemetry.md`) and the app shows the Thinking switch
-  disabled with the reason, instead of leaving a control that does nothing.
+  been dropped. The engine now renders the template at load and reports what it found as `think_ctl`
+  on `BMOE_READY` (see `docs/telemetry.md`). Where the flag is inert but reasoning is a *structural*
+  section of the format, the turn now starts past that section, built by llama.cpp's own
+  continuation hook so every family's markers come from upstream rather than from this engine.
+  Where the model owns its reasoning span and simply cannot be asked to skip it — LFM2.5 — that is
+  reported instead of papered over, and the app shows the Thinking switch disabled with the reason.
+  Measured, not assumed: handing LFM2.5 a pre-closed empty reasoning span makes it reason *untagged
+  into the answer*, worse than leaving the setting alone, so the engine does not do it.
 
 ### Changed
 - **The harmony/gpt-oss marker strings are gone from the decode path.** Priming gpt-oss to answer
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,24 @@ Semantic Versioning.
 
 ## [Unreleased]
 
+### Fixed
+- **Thinking off no longer silently does nothing** (#82). Turning Thinking off set the template
+  variable `enable_thinking` and stopped there — but that variable is only a *request* to the
+  model's chat template, and many templates never read it. LFM2.5's is one: the rendered prompt came
+  out byte-identical either way, the model reasoned on, and nothing reported that the setting had
+  been dropped. The engine now renders the template at load to find out which case a model is in,
+  and where the flag is inert it closes the reasoning span in the prompt instead, using llama.cpp's
+  own continuation hook so the markers for each family come from upstream's per-template handler
+  rather than from this engine. Models that support neither are reported as such
+  (`think_ctl` on `BMOE_READY`, see `docs/telemetry.md`) and the app shows the Thinking switch
+  disabled with the reason, instead of leaving a control that does nothing.
+
 ### Changed
+- **The harmony/gpt-oss marker strings are gone from the decode path.** Priming gpt-oss to answer
+  without reasoning used to be a literal `<|start|>assistant` suffix test and a literal
+  `<|channel|>final<|message|>` appended in `session.cpp`. It is now the same generic mechanism as
+  every other family, so the engine names no model's markers and a submodule bump that changes them
+  needs no engine change.
 - **README front page reworked around the result.** A copyable result banner under the title, an
   autoplay hero GIF of gpt-oss-120b generating on-device (real time, cut from the existing demo
   footage), and a "Try it on your phone" quickstart that starts from the release APK and the
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -203,8 +203,11 @@ static int run_session_loop(const RunConfig & cfg, IMetricsSink * sink, IRouteTr
         std::fflush(stdout);
         return 1;
     }
-    std::printf("BMOE_READY {\"load_s\":%.3f,\"arch\":\"%s\",\"n_ctx\":%d}\n", session->load_seconds(),
-                json_escape(session->arch()).c_str(), session->n_ctx());
+    // think_ctl states, once, whether this model can honour a think=false request at all, so a UI
+    // can disable its Thinking control instead of leaving one that silently does nothing (#82).
+    std::printf("BMOE_READY {\"load_s\":%.3f,\"arch\":\"%s\",\"n_ctx\":%d,\"think_ctl\":\"%s\"}\n",
+                session->load_seconds(), json_escape(session->arch()).c_str(), session->n_ctx(),
+                bmoe::think_control_name(session->think_control()));
     std::fflush(stdout);
 
     std::mutex mtx;
--- a/core/CMakeLists.txt
+++ b/core/CMakeLists.txt
@@ -25,6 +25,7 @@ if(BMOE_HAVE_LLAMA)
         src/moe/router_hook.cpp
         src/engine/session.cpp
         src/engine/chat_parse.cpp
+        src/engine/thinking_control.cpp
         src/engine/runtime.cpp
         src/metrics/csv_metrics_sink.cpp
         src/metrics/route_trace_sink.cpp
--- a/core/include/bmoe/session.h
+++ b/core/include/bmoe/session.h
@@ -53,6 +53,22 @@ struct SessionConfig {
 // remembering to touch both. n_batch = n_ctx so any prompt that fits the context prefills in one batch.
 SessionConfig session_config_from(const RunConfig & cfg);
 
+// How a GenerateRequest::think=false request can be honoured on THIS model. Decided once at
+// open() by rendering the model's own chat template, never from a list of model names.
+//
+// Turning thinking off is a request to the template, and a template is free to ignore it: the
+// flag reaches the jinja context either way, so a model whose template never reads it reasons on
+// regardless and nothing reports that the setting was dropped. Probing says which case a model is
+// in, so a caller can stop offering a control that does nothing.
+enum class ThinkControl {
+    Template, // the template reads enable_thinking — passing the flag is the whole mechanism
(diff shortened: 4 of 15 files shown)


Helldez:
## On-device results: the harmony cleanup holds, the LFM2.5 fix does not

Measured with the branch build, one prompt that needs no reasoning ("What is 2+2?"), `--chatml
--no-think`. Marking this draft: the headline claim above is **not** supported for LFM2.5.

### The probe is correct on real ggufs

Classification comes from the model actually loaded, not from the bundled `.jinja` used in the host
test, and it is right in all three cases:

| model | `think_ctl` |
|---|---|
| LFM2.5-8B-A1B Q5_K_M | `prefill` |
| gpt-oss-120b Q4_K_M | `prefill` |
| Qwen3.6-35B-A3B Q4_K_M | `template` |

### gpt-oss: the hardcoded harmony markers are safely retired

```
What is 2+2?**Answer:**
\(2 + 2 = 4\)

generation: 14 tokens
```

14 tokens, no analysis, straight to the answer — the same behaviour the two literal marker strings
used to produce, now via the generic path. This half of the PR does what it claims.

### LFM2.5: the prefill does not suppress reasoning

500 tokens requested, and it never reaches an answer:

```
What is 2+2?
Okay, so I need to figure out what 2 plus 2 equals. Let me start by recalling basic arithmetic.
[... 500 tokens of reasoning, no answer, no tags ...]
```

The prompt ends with the reasoning span already closed (`<|im_start|>assistant\n<think>\n\n</think>`,
verified in the host test). The model simply reasons past it, in the **answer**, untagged.

**This is the #83 failure reached by a different route.** #83 forced the block closed on logits and
the model reopened it, then abandoned the tags; here the block is closed before generation and the
model never opens one at all — same end state, reasoning sitting in the reply instead of a
collapsible block. Worse than the original bug, where at least the reasoning parsed into a block.

### What this means

The continuation hook proves only that the *renderer* can produce a closed span. It says nothing
about whether the *model* was trained to treat one as "reasoning already done". Qwen3 was — its own
template emits exactly this. LFM2.5 was not, and Liquid documents no way to disable reasoning on it
at all; they ship a separate non-reasoning checkpoint instead.

So `prefill` currently conflates two different things:

- **a channel switch** (harmony): the model is structurally placed in the answer section — works.
- **an empty closed thought** (LFM2.5): a convention the model has to have been trained on — does not.

There is no way to tell those apart from the template alone without reintroducing per-family
knowledge, which is what this PR exists to remove.

### Open question, not being decided unilaterally

#83 recorded that injecting a short concluding *line* instead of closing on nothing changed LFM2.5's
behaviour completely (16 tokens instead of 166). That signal is real and reproducible, but the line
was invented English tuned against one model — rejected then, and no more shippable as an engine
constant now. Whether the honest form of that is caller-supplied configuration (engine chooses
nothing, default empty), or whether a model that ignores a closed span should simply be reported as
`none`, is the decision this PR is now blocked on.

Helldez:
## Ruled out: this is not the known LFM2.5 template-detection bug

Worth checking, since llama.cpp does carry LFM2.5 bugs in exactly this area — most relevantly
[#23852](https://github.com/ggml-org/llama.cpp/issues/23852), where LFM2.5 fails detection and falls
back to a content-only parser, making `<think>` tags visible. That is a different failure with a
similar smell, and it does **not** apply here:

- **The pin has the fix.** The submodule base is 2026-07-10, after #23852's fix (#24178), and the
  dedicated LFM2.5 detection branch is present (`common/chat.cpp:2600`).
- **Detection actually succeeds on this gguf.** #23852's cause is that the template builds its tool
  list as `"List of tools: " + (tools | tojson)`, so the literal `List of tools: [` never appears in
  the source. In the template shipped inside this gguf it *does* appear, at line 54.
- **The template in the gguf is the one under test.** Extracted from the model file on the device and
  compared byte-for-byte with `models/templates/LFM2.5-8B-A1B.jinja`: identical once line endings are
  normalised. So the host test above is rendering exactly what the device renders.
- **The rendered prefill has the LFM2 handler's own shape** (`<|im_start|>assistant\n<think>\n\n</think>`),
  which is only produced by the branch at `common/chat.cpp:1713-1722`. The model was routed correctly.

So the prompt was well-formed, correctly routed, and the reasoning span was genuinely closed before
generation. LFM2.5 reasons past it anyway.

On suppression specifically, upstream's position is
[#20196](https://github.com/ggml-org/llama.cpp/issues/20196) — same class of failure, closed **not
planned**. And Liquid's own framing is that LFM2.5 is a reasoning model whose template carries no off
switch; they ship a separate non-reasoning checkpoint instead. This is the model's design, not a bug
waiting on an upstream patch.

Helldez:
## Resolved: the prefill now applies only where the model cannot decline it

The regression above is fixed, and the fix is the distinction the earlier probe was missing.

**gpt-oss declares no reasoning tags** — reasoning is a *channel*, a section the format itself
separates, so starting the turn past it is not something the model can refuse. **LFM2.5 declares
`<think>`/`</think>`** — the span is the model's own to open and close, so handing it one already
closed and empty is a *suggestion*, and this model was not trained on that convention. Liquid ships
a separate non-reasoning checkpoint rather than an off switch.

So the probe reads that off `common_chat_params::thinking_start_tag`/`thinking_end_tag` — model-
supplied data, no names, no template string matching. Flag inert **and** tags declared → `none`:
report it, disable the control, say why. Flag inert and no tags → the prefill binds.

### Measured on-device, all three catalog families

| model | `think_ctl` | Thinking off produces |
|---|---|---|
| LFM2.5-8B-A1B Q5_K_M | `none` | reasoning in `reasoning_content`, answer clean (`4`) — i.e. `main`'s behaviour, now honestly labelled |
| gpt-oss-120b Q4_K_M | `prefill` | `"reasoning":""`, direct answer — what the retired hardcoded markers produced |
| Qwen3.6-35B-A3B Q4_K_M | `template` | `"reasoning":""`, direct answer — untouched path |

LFM2.5 is no longer made worse: with `none` the engine does exactly what `main` does, so the 500-token
untagged-prose failure is gone and the reasoning is separable again. What changes for it is that the
app now disables the Thinking switch and explains why, instead of leaving one that does nothing.

Host: 7/7 ctest (MoE byte-identity gates included), clang-format 18, arm64 cross-build, Kotlin — all
clean. CI still blocked on Actions billing, so this is all local.

Ready for review.

PR #90, opened 2026-07-20
feat: contiguous per-expert sidecar (+16% decode measured on device)

Helldez:
## What

A contiguous per-expert **sidecar file** (`<model>.experts.bmoe`): an offline byte re-ordering of the gguf's expert tensors so one routed expert costs **one contiguous O_DIRECT read** instead of one pread per projection. Same weights, same quantization, byte-identical output. The gguf stays untouched (llama.cpp still loads it; no repack, no fork) - only the streamer's expert reads move to the sidecar.

- `bmoe-cli --build-sidecar` (one-shot builder, recipe-driven discovery, no model constants) and `--expert-sidecar PATH|auto`
- Library API `bmoe/expert_sidecar.h` so the app can build it post-download later
- `FileReader::read_scatter`: one aligned window, N destinations
- `bmoe-iobench --scatter N`: the microbench that justified it
- Identity check (source size + expert-map FNV hash): a stale or tampered sidecar is a **hard init error**, never a silent fallback (gate G8e)

## Why

`--scatter` microbench on the reference device: at the split-layout slice size (3 x ~300 KiB per expert) scattered reads plateau ~10% under the flash ceiling even at 16 lanes, and cost ~35% at the engine's real ~2 effective lanes (1355 vs 1826 MiB/s). The roadmap's expert-contiguous item asked to be re-justified before being built; this is that justification, measured.

## Measured

On-device A/B (LFM2.5-8B-A1B, interleaved A/B/A/B, condition-gated cooldown): **+15.7% decode tok/s** (2.529 -> 2.927), entirely attributed to io_ms (327 -> 282 ms). Raw CSVs + run log in `docs/bench-data/2026-07-20-sidecar/`.

## Gates

G8a-d: streamed == resident byte-identity with the sidecar across cache off / LRU / overlap / prefetch, on both expert layouts (qwen3moe split, gemma4 fused). G8e: tampered sidecar refused. Full ctest suite passes on the host.

## Owed / follow-up

- Qwen3-30B-A3B k=8 A/B (pure I/O-bound; microbench predicts a larger gain)
- App wiring: build post-download + Settings toggle (default off until measured there)
- Disk cost is ~the model's expert bytes; documented in the README feature entry


Helldez:
Closing unmerged: measured negative in the shipping configurations. Three interleaved on-device A/Bs: +15.7% serial cache-off (LFM2.5), but -20% with overlap+cache 3000 (LFM2.5) and -23% at the README config on Qwen3-30B k=8. The microbench premise (contiguity buys bandwidth) is true; the engine conclusion is not - the overlap kernel consumes projection-major, and a whole-entry read triples latency-to-first-slice, stalling the pipeline. Full analysis and raw CSVs: docs/bench-data/2026-07-20-sidecar/findings.md (merged via #91). Implementation preserved at tag expert-sidecar-refuted; a measured regression does not merge.

PR #101, opened 2026-07-23
feat(moe): measure how predictable expert routing is, then act on it — and refute it

Helldez:
## Why

[Temporal prefetch](docs/prefetch.md) was built on a predictor nobody had priced. Measured after the
fact over the committed route traces, the previous-token bet is right about **38% on Qwen and 18% on
gpt-oss** — not enough to pay for the reads it speculates. The lesson was not that prefetch is
impossible; it was that a predictor should be measured *before* it is wired into anything.

So this PR does both halves in order: it adds the instrument, measures a much better predictor with
it, acts on that predictor — and then reports that acting on it **loses**, which is why the lever
ships off by default.

## `--predict-log`: the instrument

It ranks each layer's experts a layer early, by running the **next** layer's router matrix on the
**current** layer's gate input. The residual stream barely moves between layers, so the stale input
ranks nearly as the real one will — the mechanism [FATE](https://arxiv.org/html/2502.12224v1)
reports 78.8% for. Training-free and model-unchanged: the router matrix is a dense weight already
resident, and the prediction is one GEMV per layer.

The matrix is **learned from the graph** (the gate matmul's first source) rather than looked up by
tensor name, so no architecture is named anywhere in the path. Being a weight leaf, its pointer
stays valid into layers the current token has not reached — which is what makes predicting forward
possible at all.

Three predictors, scored on the same routings of the same run:

| | what it is |
|---|---|
| `stale` | the next layer's router run a layer early — the one under test |
| `prev` | what the previous token routed here — the bet `--prefetch` already places |
| `ctrl` | the same code with the staleness removed — **a control, not a predictor** |

**The control is the load-bearing part.** It shares every line of code with the prediction under
test and differs only in using the layer's own matrix, so it *must* reproduce the selection
llama.cpp computes from those same two tensors. A transposed matrix, a mis-strided row or the wrong
token of the batch collapses it toward chance while the stale figure would stay superficially
plausible. An architecture that selects by something other than raw-logit ranking (an additive bias,
group-limited routing) puts it below 100%, and by that much the stale figure understates the method
— the CLI says so rather than letting the gap be blamed on staleness.

Per layer as well as aggregate, because an aggregate flatters a prefetch: what a prefetch costs is
set by the layers it gets **wrong**, and a MoE model's first layers route far less predictably than
its last. Denominators are printed per predictor — the stale gate structurally cannot rank layer 0
(nothing precedes it) or the first token of a run, and those are counted as *unscored* rather than
folded in.

## What it measured

On device, against what `--prefetch` already bets on:

| model | stale-gate | prev-token | control |
|---|---|---|---|
| Qwen3-30B-A3B (top-8 of 128) | **88.6%** | 43.3% | 100.0% |
| Qwen3.6-35B-A3B (top-8 of 256) | **80.7%** | 35.4% | 100.0% |

The control at exactly 100% on every layer of both says Qwen selects by pure logit ranking, so
those numbers need no discount. The prev-token column corroborates the offline route-trace
estimates that motivated the whole exercise.

## `--predict-prefetch`: acting on it, and the verdict

The prediction is fed to the same speculative read path `--prefetch` uses (same cache buffers,
accounting and summary line, tagged `[stale-gate]`), with two design points that matter more than
the swap: speculation is **issued after the current layer's load** (every load path begins by
quiescing speculation, so an earlier queue would be cancelled), and it is **drop-aware** — with
`--drop-cold-experts` armed, a predicted expert below the drop threshold is not speculated, because
if it misses the policy would discard it unread. `--predict-spec-max N` bounds how much flash the
prediction may spend per layer; `0` is retention-only, where predicted residents are LRU-protected
(new `IExpertSource::retain`) and nothing is read ahead.

It was then measured across the whole cost spectrum, and **the read-ahead loses**:

- **−21%** against the shipping drop + pinned-dense baseline at `--predict-spec-max 2`, in the one
  thermally matched adjacent pair of the campaign — with hit rate **up** 4.3pp, 79% of speculations
  useful, and +37% flash bytes per token.
- **−28%** on 8 I/O lanes in the other matched pair, effective flash bandwidth collapsing 585 →
  392 MiB/s — consistent with the measured lane ceiling.
- **Retention is hit-rate-neutral** (77.2% vs 77.6% at a 3000 MiB cache), exactly as the offline
  replay bound predicted: within a two-layer horizon, eviction order is irrelevant at that size.

Speculation improved every metric it owns and still lost the wall clock, by the same mechanism that
killed more lanes, coalescing and the expert sidecar: on a saturated flash, a better guess spends
the bandwidth that was the binding constraint. So the flag ships **off by default** and the app
defaults its budget to retention-only.

A method note worth keeping: the first day's comparisons were **invalidated by silent thermal
capping** — re-running the baseline on a heated device reproduced byte-for-byte identical I/O, hit
rate and drop counts at −40% wall clock. Only thermally matched adjacent cells are quoted above;
`docs/expert-prediction.md` records the contaminated numbers too, and why they cannot be read.

Two structural findings survived the campaign and are in the code: the observer tax was 109 ms/token
before the rebuild — ~35-45 ms per GEMV pass from calling `ggml_fp16_to_fp32` per element (21M calls
per token) plus ~20 ms of barrier — and the prefetch's barrier-less gate-input read is guarded by a
sampled **self-validating watchdog** that disarms the predictor out loud if a future llama.cpp
planner change makes the read untrustworthy.

## Cost and scope

- **Byte-identical**: the probe reaches neither `load_layer`, nor the cache, nor the graph (`G9a`);
  the prefetch only warms the cache (`G10a`), with the same documented exception as `--prefetch`
  under `--drop-cold-experts`, where a correct guess un-drops an expert and buys quality, not speed.
- **The probe is not free**: one isolated node and two GEMVs per layer on the eval thread. A probed
  run is **not a benchmark run**; the docs and `--help` both say so.
- Decode only; requires `--moe-stream`. `--predict-prefetch` needs the LRU cache and excludes
  `--prefetch`. Routings wider than 32 experts are declined rather than scored against a truncated
  ranking.

## Verification

- `ctest` — 7/7 pass, including `G9a`/`G9b` (probe inert; control 100.0% over 96 routings) and
  `G10a`/`G10b` (speculation inert; 33 experts actually speculated, so the identity is not vacuous)
  on both tiny models (qwen3moe, gemma4).
- CI format gate (`clang-format` 18) clean.
- On-device measurement is reported above, from thermally matched pairs only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,7 +80,10 @@ Semantic Versioning.
   a dedicated worker at an L+2 horizon (the probe's stale-2 column prices that staleness per
   model). Throughput verdict OPEN pending thermally matched A/B pairs — the first day's
   comparisons were invalidated by silent thermal capping (see docs/expert-prediction.md, which
-  also records the one matched pair measured: speculation on 8 io lanes loses −28%).
+  also records the one matched pair measured: speculation on 8 io lanes loses −28%). The example
+  app exposes it as an experimental toggle in the Streaming section (off by default, mutually
+  exclusive with the temporal-prefetch setting, spec-max rungs 0/1/2/4) so the matched pairs can
+  be run from the app itself.
 
 ### Fixed
 - The 0.15.0 entry for the app default still called the quality cost unquantified, contradicting the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,29 +21,6 @@ Semantic Versioning.
   reservation pushing an already-tight system into reclaim — nearly 6 000 major faults per token,
   none of them the workload's fault.
 
-### Fixed
-- A stray `%` in the `--dense-weights` help text was read by `printf` as a conversion specifier, so
-  that line printed garbage and read past the argument list.
-- `scripts/build-android.ps1` now judges `cmake` by its exit code instead of by whether it wrote to
-  stderr, so it works under Windows PowerShell 5.1 and not only under `pwsh` (the NDK toolchain
-  prints progress to stderr, which 5.1 turns into a terminating error).
-
-## [0.15.1] - 2026-07-23
-
-### Added
-- **A warning when cache-aware dropping meets a narrow routing.** The threshold is a fraction of the
-  uniform share `1/top-k`, so `--drop-cold-experts 0.75` means "skip below 9.4% of the routing" at
-  top-k 8 — where every number in 0.15.0 was measured — but "below 18.8%" at top-k 4 and "below
-  37.5%" at top-k 2, where a miss discards the whole minority expert. The engine now says so once at
-  load when the effective top-k is 4 or fewer, quoting the real share for the model in hand, and the
-  app shows the same caveat inline under the setting. **gpt-oss is the case this exists for**: it
-  routes 4 of 128, and the app default is 75%, so 0.15.0 shipped that combination with nothing
-  saying it was outside the measured range.
-  It warns rather than clamping: the engine cannot know whether the trade is acceptable for a given
-  model, and silently adjusting a number the caller chose would be worse than a loud caveat.
-- `BMOE_READY` gains `n_expert_used`, the effective routing width after any override (0 on a non-MoE
-  model), so a UI can interpret the setting at all; `Session::n_expert_used()` exposes the same to
-  embedders. Additive — older consumers ignore it.
 - **`--predict-log` — measure how predictable the routing is, without acting on it.** For every
   decoded token the engine ranks each layer's experts a layer early, by running the **next** layer's
   router matrix on the **current** layer's gate input — the residual stream barely moves between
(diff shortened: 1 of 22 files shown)


Helldez:
## Status after the 2026-07-23/24 measurement campaigns

**The PR contains three things**: the probe (`--predict-log`, unchanged contract, plus a stale-2
column), the actuation (`--predict-prefetch` + `--predict-spec-max N`, off by default), and the
measured record of two days of on-device A/B (`docs/expert-prediction.md`).

### Accuracy — proven

- Stale-gate: **88.6%** of routed slots on a 48-layer/128-expert model, **80.7%** on a
  40-layer/256-expert one; zero-staleness control at exactly **100.0%** on every layer of both, so
  the numbers need no discount. Prev-token predictor: 43/35% — matching the offline replay
  estimates.
- The observer tax is gone: the first implementation cost 109 ms/token (a per-element
  exported-function F16 conversion — 21M calls/token — plus a barrier per layer). Now: native
  `__fp16` GEMV (~40×), no barrier (ask-pass pointer capture validated by a sampled watchdog that
  never tripped), prediction on a worker at an L+2 horizon, `retain()` for predicted residents.

### Throughput — settled by thermally matched pairs

The first day's pairwise comparisons were withdrawn: every variant ran on a progressively hotter
device against a cool-morning baseline, and re-running the baseline itself at the end of the day
gave 3.93 tok/s vs the morning's 6.54 with byte-identical I/O, hit and drop counts — deterministic
engine, silent thermal capping. The redo used same-session pairs with a sentinel replicate.

- **Read-ahead loses.** Spec-max 2 vs baseline, adjacent cells at the same clock caps and the same
  battery temperature: **3.14 vs 3.96 tok/s (−21%)** — with hit rate *up* 4.3pp, 79% of
  speculations useful, +37% flash bytes/token. Same verdict on 8 io lanes (−28%). The mechanism is
  the engine's recurring one: the flash is saturated, so a speculative read adds no bandwidth — it
  competes with the demand reads that were the constraint. Every intermediate metric improves; the
  wall clock pays.
- **Retention is inert at a 3000 MiB cache**: hit rate moved 0.1pp / 0.4pp across the two days —
  the offline replay bound (~5pp ceiling for any policy over LRU) confirmed from inside the engine.
  With no hit-rate movement there is no mechanism left for it to win by, only the residual
  gate-GEMV cost.

### Disposition

The probe and its record stay — the instrument is the deliverable, and the accuracy result is
real and reusable. The actuation stays off by default; the example app now defaults its spec-max
rung to **0 (retention only)**, the one configuration the matched pairs did not refute, and its
settings copy states the measured verdict instead of an open question. Separately, the measured
levers toward the throughput target (k=6, larger cache) need nothing from this PR.

PR #104, opened 2026-07-26
feat(gpu): --gpu — run the dense half on the GPU, and the measurement that says not to

Helldez:
Runs the **dense (non-expert) half** of the model on a GPU, and reports — with numbers — that on the hardware this engine targets it does not pay. Ships gated and off, because the measurement is worth having and the feature is not worth recommending.

> **Verdict up front.** On an integrated mobile GPU, decode is slower at every placement measured: full dense offload **0.73×**, output head alone **0.80×**, and (in a companion experiment) routed experts held in device memory **0.22×**, with even a fully resident model and nothing streamed at **~0.79×**. Output is token-identical throughout, so this is a performance result and never a correctness one. The GPU's one unrefuted advantage is batched **prefill** (~2.3×), which is not measured here.

## Why the experts stay on the CPU

The streamer rebinds an expert tensor's `->data` to a host buffer it refills from flash before every token. No GPU backend follows that: ggml's OpenCL backend reads `tensor->data` exactly once, in `init_tensor` at load, and dispatches against `cl_mem` handles cached in `tensor->extra` thereafter — the pointer it read is not even a real address. There is no zero-copy path either (no `CL_MEM_USE_HOST_PTR`, no SVM, no dma-buf import), so a device-side weight is a *copy*. A streamed expert placed there would silently compute whatever the warm-up decode left behind.

That argument is about rebinding a **pointer**, and it holds. It does not rule out rebinding the *contents*, which is what the companion experiment does — see the bottom of this description.

## Why the dense half looked worth moving

A MoE uses **100% of its dense weights on every token** but only `k/n_expert` of its expert weights, which puts the halves in the same order of magnitude:

| Model | dense | expert bytes touched/token |
|---|---:|---:|
| Qwen3-30B-A3B (128 experts, top-8) | 951 MiB | ~1.10 GiB |
| Qwen3.6-35B-A3B (256 experts, top-8) | ~1 GiB | ~0.67 GiB |

Decode here is compute-bound once the I/O levers are in, so moving ~half the arithmetic off four phone cores looked like the largest remaining lever that costs no quality. It is **lossless** — placement changes where a matmul runs, not what is computed.

That reasoning was about **how much work** the dense half is. It said nothing about **where that work sits in the graph**, and that turns out to be what decides it.

## Design

Two things set together before the load:

- `llama_model_params::devices` gets the discovered GPU; `n_gpu_layers` says how much (`-1` = all).
- `tensor_buft_overrides` pins the expert tensors to the CPU buffer type. llama.cpp resolves a matching override **before** the layer's device assignment, so an expert stays in host memory even in an offloaded layer.

**Device-agnostic.** Discovery goes through the ggml backend registry (`core/src/engine/gpu_device.h`) and names no vendor or API. A build with no GPU backend and a device with no driver produce the same result — no device found — and the engine falls back to the CPU with a note. `--gpu-require` turns that into a failure, so an A/B cannot silently compare CPU against CPU.

**Model-agnostic.** The pin pattern is derived from the architecture recipe (`expert_tensor_pattern()`), so a new MoE family added as a recipe row pins its own experts with nothing else to update. A unit test asserts every registered recipe yields a pattern.

## `--ubatch N`, the lever that did pay

Compute buffers are reserved for the widest graph, and the engine had always set `n_ubatch = n_ctx` so any fitting prompt prefills in one pass — which quietly tied resident memory to the **context** rather than to the work:

| n_ctx | OpenCL compute buffer | CPU compute buffer |
|---|---:|---:|
| 2048 | 1203 MiB | 320 MiB |
| 512 | 301 MiB | 80 MiB |

Exactly 4× for a 4× context, on both backends — nothing to do with the GPU. On an engine whose whole problem is that the expert cache and the dense weights compete for RAM, that was a real budget and it was invisible. Decode is unaffected (a decode graph is one token wide regardless); the cost is prefill throughput. Default 0 keeps the previous behaviour exactly.

This is also what invalidated the first measured verdict, below.

## What it measures

Three passes are recorded rather than one, because the first two were wrong in instructive ways and the corrections are the useful part.

**Pass 1 said 2.6× slower.** Wrong: that was our own compute-buffer reservation (~900 MiB of it) pushing an already-tight system into reclaim — 5 958 major faults per token. Capping `--ubatch` took the same cell from 1.207 to 2.752 tok/s.

**Pass 2 said `--gpu-layers 1` was worth +7%.** Also wrong: a repeat reversed the sign. This device's run-to-run noise is ±6%, larger than every GPU effect measured except the full-offload penalty.

**Pass 3, on a second model with a faster CPU baseline, is unambiguous.** `Qwen3.6-35B-A3B` (Q4_0, 40 layers) with the current recipe including `--drop-cold-experts`, run as a palindrome, `scaling_max_freq` logged per cell because the SoC changed clock state mid-sweep:

| `--gpu-layers` | clock | tok/s | compute s/tok | majflt/token |
|---|---|---:|---:|---:|
| 1 (output head only) | 3.03 GHz | 5.064 | 0.141 | 7.11 |
| **0 (CPU)** | 3.03 GHz | **6.383** | 0.098 | 0.29 |
| **0 (CPU)** | 2.27 GHz | **5.528** | 0.131 | 0.49 |
| 1 (output head only) | 2.27 GHz | 4.663 | 0.157 | 1.65 |
| all 40 layers | 2.27 GHz | 4.062 | 0.192 | 9.95 |

Full offload **−27%**, output head alone **−16 to −21%**, same sign in both halves of the palindrome. This does not contradict the neutral result on Qwen3-30B; it explains it. The CPU baseline got faster (`--drop-cold-experts` is worth ~+18% on this model), so the same fixed cost is a larger share of a smaller total.

## Why it loses, mechanically

The dense and expert halves **interleave**: a MoE layer is dense → routed experts → dense → routed experts, once per layer. Splitting them across two devices crosses the boundary **twice per layer** — 96 graph splits against 1 — and every crossing copies and synchronises an activation.

The GPU genuinely frees the CPU: occupancy falls 88% → 54% at full offload, and the overlapped I/O residual even improves (0.035 → 0.030 s/token). But compute rises 0.131 → 0.192 and ~10 major faults per token appear from the compute buffer's memory pressure. The freed CPU time is real; the boundary tax eats all of it.

`--gpu-layers 1` comes closest because the output head is the one dense component with no expert layer after it to hand control back to — 2 splits instead of 96. It is still not a win: one matmul is too small a share of the per-token work.

**And the deeper reason is the one this PR's original description guessed at and then set aside:** batch-1 decode is GEMV, memory-bound, and an integrated GPU shares the same LPDDR as the CPU, so there is no bandwidth to win. That guess was right, and it outranks the byte-count argument that motivated the feature.

## Flags and reporting

`--gpu`, `--gpu-layers N`, `--gpu-require`, `--ubatch N`. Reporting is **resolved, never requested**: `BMOE_READY` gains `"gpu"` and the CSV preamble gains `gpu=`, empty/`cpu` when the dense path ran on the CPU, so a bench cell cannot be mistaken for a GPU cell it silently was not. The app's **Dense weights on GPU** toggle displays the device the engine actually got and warns on a fallback.

For benchmarking, compare toggle-on against toggle-off **on the same binary**: an OpenCL-enabled build registers and initialises the backend at startup whether or not the toggle is on.

## Build

Off by default at every level. `-DBMOE_OPENCL=ON` is opt-in because linking the backend makes `libOpenCL` a hard load-time dependency: such a binary will not start on a device with no OpenCL driver, where the default build runs everywhere.

```
pwsh scripts/fetch-opencl-android.ps1    # Khronos headers + an ICD loader to link against
pwsh scripts/build-android.ps1 -OpenCL
```

The loader is a **link stub only** and is deliberately not shipped — at runtime `libOpenCL.so` resolves to the device's own `/vendor/lib64` driver. Shipping the Khronos loader would outrank the vendor driver and then find no ICD behind it: everything links, nothing crashes, no GPU is ever found. The app also needs `<uses-native-library android:name="libOpenCL.so" android:required="false"/>`, without which the linker kills the process before `main`.

`scripts/build-host.sh` and CI pass no GGML flags, so the default host build and the byte-identity gates are untouched.

## Also fixed

- `--dense-weights anon|ahwb` no longer tries to rebind tensors the GPU has taken (they are not under host-memory reclaim pressure; rebinding a device-resident tensor would be ignored at best).
- A stray `%` in the `--dense-weights` help text was read by `printf` as a conversion specifier, so that line printed garbage and read past the argument list. Pre-existing, unrelated.
- `scripts/build-android.ps1` judges `cmake` by its exit code rather than by whether it wrote to stderr, so it works under Windows PowerShell 5.1 and not only under `pwsh`.

## The companion experiment

Whether the routed **experts** can move to the GPU at all is answered separately, and negatively, on [`feat/gpu-expert-slots`](https://github.com/Helldez/BigMoeOnEdge/tree/feat/gpu-expert-slots): they can — by rebinding contents rather than a pointer, into a flash-filled residency pool — it is bit-exact, and it measures 0.22×. That work depends on seams carried on a llama.cpp fork branch, so it deliberately stays out of this tree and out of this PR.

## Verification

- Host gates **7/7** green; the streaming path is untouched.
- `config_validate` extended: the new knobs, and every recipe's expert pattern.
- clang-format 18 clean.
- Android arm64 cross-compile with `-OpenCL` succeeds; `libggml-opencl.so` carries the expected unresolved `libOpenCL.so` `DT_NEEDED` and is packaged in the dev APK.
- README benchmark tables deliberately untouched: the new cells are Q4_0 and those tables are Q4_K_M, so the numbers are not comparable.
- Full evidence and method notes in `docs/bench-data/2026-07-27-gpu-dense-offload/`, including the two wrong verdicts and why they were wrong.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 /out/
 cmake-build-*/
 
+# OpenCL headers + ICD loader, fetched by scripts/fetch-opencl-android.ps1 for the optional
+# GPU build. Third-party sources we do not vendor: they are a build prerequisite, not our code.
+/third_party/opencl-sdk/
+
 # Models and large artifacts (never commit weights)
 *.gguf
 *.bin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,14 @@ Semantic Versioning.
   is reproducible, not because it is recommended. See `docs/gpu-offload.md` and
   `docs/bench-data/2026-07-27-gpu-dense-offload/`, which also records the two wrong verdicts that
   preceded this one and why they were wrong.
+
+  Repeated on a second model (`Qwen3.6-35B-A3B`, Q4_0, with `--drop-cold-experts`) the sign is no
+  longer ambiguous: full offload **−27%**, output head alone **−16 to −21%**, both outside the noise
+  and consistent across a palindrome. The CPU baseline got faster, so the same fixed
+  boundary-crossing cost is now a larger share of a smaller total. Decode belongs to the CPU on this
+  class of device at every placement measured — including holding the routed experts in device
+  memory, which is a separate experiment on a fork-dependent branch and measures 0.22×. The GPU's
+  one unrefuted advantage is batched **prefill** (~2.3×), which is not measured here.
 - **`--ubatch N`: the widest graph computed at once, decoupled from the context.** Compute buffers
   are reserved for the worst-case graph, and the engine had always set `n_ubatch = n_ctx` so that
   any fitting prompt prefills in one pass — which quietly ties resident memory to the context
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,15 +29,17 @@ Semantic Versioning.
   because linking it makes `libOpenCL` a hard load-time dependency: such a binary will not start on
   a device with no OpenCL driver, where the default build runs fine.
 
-  **Measured on a phone: behind the CPU, and not yet settled.** 3.175 tok/s on the CPU against
-  2.752 with the dense path offloaded — a ~15% gap, once `--ubatch` below removed a self-inflicted
-  cost that had made it look like 2.6x. What remains is structural: the dense and expert halves
-  interleave at every layer, so a two-device split crosses the boundary twice per layer (98 graph
-  splits against 1) and each crossing copies an activation. The configuration that would settle it,
-  `--gpu-layers 1` (output head only, 1-2 splits instead of 98), is owed. Output is token-identical
-  throughout, so this is a performance question and never a correctness one. Ships gated and off so
-  the measurement is reproducible, not because it is recommended. See `docs/gpu-offload.md` and
-  `docs/bench-data/2026-07-27-gpu-dense-offload/`.
+  **Measured on a phone: between neutral and mildly negative, and inside the noise.** At a capped
+  ubatch, CPU reads 2.941–3.116 tok/s across cells and full offload reads 2.768; `--gpu-layers 1`
+  (output head only) read 3.029 in one pass and 2.934 in a repeat that reversed its sign. The
+  device's run-to-run variation is ±6%, larger than every effect measured except the full-offload
+  penalty. The GPU does take real work off the CPU — occupancy falls from 88% to 54% — but the dense
+  and expert halves interleave, so a two-device split crosses the boundary twice per layer (96 graph
+  splits against 1) and the boundary tax eats the saving. Output is token-identical throughout, so
+  this is a performance question and never a correctness one. Ships gated and off so the measurement
+  is reproducible, not because it is recommended. See `docs/gpu-offload.md` and
+  `docs/bench-data/2026-07-27-gpu-dense-offload/`, which also records the two wrong verdicts that
+  preceded this one and why they were wrong.
 - **`--ubatch N`: the widest graph computed at once, decoupled from the context.** Compute buffers
   are reserved for the worst-case graph, and the engine had always set `n_ubatch = n_ctx` so that
   any fitting prompt prefills in one pass — which quietly ties resident memory to the context
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,50 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [Unreleased]
+
+### Added
+- **`--gpu`: the dense half of the model can run on a GPU.** The routed experts always stay on the
+  CPU, and structurally must — the streamer rebinds their `->data` to host buffers it refills every
+  token, while a GPU backend copies weights into device memory once at load and never re-reads the
+  pointer. What a GPU *can* take is everything else: embeddings, attention/SSM, norms, `lm_head`.
+  On a MoE that is roughly half the per-token arithmetic, because 100% of the dense weights are used
+  on every token against only `k/n_expert` of the expert weights (Qwen3-30B-A3B: 951 MiB dense
+  against ~1.10 GiB of expert bytes touched). Lossless — placement changes where a matmul runs, not
+  what the model computes — and orthogonal to streaming, so it combines with every existing knob.
+  Also `--gpu-layers N` to offload only the last N layers, and `--gpu-require` to fail rather than
+  fall back, so a benchmark cannot silently compare CPU against CPU.
+
+  Two properties it was built to have. It is **device-agnostic**: availability is probed at load
+  through the ggml backend registry (`core/src/engine/gpu_device.h`), which names no vendor or API,
+  so a build without a GPU backend and a device without a driver take the same clean CPU fallback.
+  And it is **model-agnostic**: the buffer-type override that pins the experts is derived from the
+  architecture recipe (`expert_tensor_pattern()`), so a new MoE family added as a recipe row pins
+  its own experts with nothing else to update.
+
+  Off by default at every level. The OpenCL backend is opt-in at build time (`-DBMOE_OPENCL=ON`)
+  because linking it makes `libOpenCL` a hard load-time dependency: such a binary will not start on
+  a device with no OpenCL driver, where the default build runs fine. See `docs/gpu-offload.md` for
+  the build recipe and, importantly, for what is **not** yet known — this ships gated and measured
+  on nothing; the argument for it is the byte split plus the compute/stall attribution, not a
+  benchmark.
+- `BMOE_READY` gains `gpu`, and the metrics CSV preamble gains `gpu=`, both carrying the **resolved**
+  device (empty / `cpu` when the dense path ran on the CPU). A run that asked for the GPU on a
+  device without one records the fallback, so a bench cell can never be mistaken for a GPU cell it
+  silently was not. The example app exposes the whole thing as a **Dense weights on GPU** toggle in
+  the Compute section, which displays the device the engine actually got rather than echoing the
+  switch back.
+
+### Fixed
+- `--dense-weights anon|ahwb` no longer tries to rebind tensors the GPU has taken; they are not
+  under host-memory reclaim pressure, so skipping them loses nothing, and rebinding a device-resident
+  tensor would have been ignored at best.
+- A stray `%` in the `--dense-weights` help text was read by `printf` as a conversion specifier, so
+  that line printed garbage and read past the argument list. Pre-existing, unrelated to the above.
+- `scripts/build-android.ps1` now judges `cmake` by its exit code instead of by whether it wrote to
+  stderr, so it works under Windows PowerShell 5.1 and not only under `pwsh` (the NDK toolchain
+  prints progress to stderr, which 5.1 turns into a terminating error).
+
 ## [0.15.1] - 2026-07-23
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,10 +27,17 @@ Semantic Versioning.
 
   Off by default at every level. The OpenCL backend is opt-in at build time (`-DBMOE_OPENCL=ON`)
   because linking it makes `libOpenCL` a hard load-time dependency: such a binary will not start on
-  a device with no OpenCL driver, where the default build runs fine. See `docs/gpu-offload.md` for
-  the build recipe and, importantly, for what is **not** yet known — this ships gated and measured
-  on nothing; the argument for it is the byte split plus the compute/stall attribution, not a
-  benchmark.
+  a device with no OpenCL driver, where the default build runs fine.
+
+  **Measured, and on a phone it loses: 2.6x slower than the CPU**, degrading monotonically with how
+  much is offloaded (3.175 tok/s CPU, 2.985 at `--gpu-layers 12`, 1.207 fully offloaded). Two costs
+  compound, both scaling with the offload: the dense and expert halves interleave at every layer, so
+  a two-device split crosses the boundary twice per layer (98 graph splits against 1); and the GPU
+  shares the same LPDDR, so its ~2 GiB of allocations add to memory pressure rather than relieving
+  it, producing 5 958 major faults per token against zero on the CPU run. Output was
+  token-identical, so this is a performance verdict and not a correctness one. The feature ships
+  gated and off so the measurement is reproducible, not because it is recommended. See
+  `docs/gpu-offload.md` and `docs/bench-data/2026-07-27-gpu-dense-offload/`.
 - `BMOE_READY` gains `gpu`, and the metrics CSV preamble gains `gpu=`, both carrying the **resolved**
   device (empty / `cpu` when the dense path ran on the CPU). A run that asked for the GPU on a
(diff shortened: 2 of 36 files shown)


Helldez:
## Measured on device: refuted — 2.6x slower than the CPU

Matched pair, `Qwen3-30B-A3B-Q4_K_M`, streaming + cache 2000 + overlap + `--dense-weights anon`, GPU cell run **first** so the thermal advantage went to it:

| Config | tok/s | compute s/tok | majflt/token | graph splits |
|---|---:|---:|---:|---:|
| CPU (no offload) | **3.175** | 0.124 | **0** | 1 |
| `--gpu-layers 12` | 2.985 | — | 234 | 24 |
| `--gpu` (all layers) | **1.207** | 0.614 | **5 958** | 98 |

Monotone in how much is offloaded, and flash traffic is identical across cells (485.5 vs 486.0 MiB/token) — the streamer did the same work in each.

**Two costs compound, both scaling with the offload.** The dense and expert halves interleave — a MoE layer is dense → experts → dense → experts, 48 times — so a two-device split crosses the boundary twice per layer: 98 graph splits against 1, each copying and synchronising an activation. And the GPU has no memory of its own: it shares the same LPDDR, so offloading *adds* memory pressure rather than relieving it (785 MiB of weights plus a fixed 1203 MiB compute buffer that `--gpu-layers` does not shrink). That reintroduces the exact fault storm `--dense-weights anon` exists to end — **5 958 major faults per token against zero**.

**The argument this PR was built on was true and irrelevant.** "The dense half is ~half the per-token arithmetic, and decode is compute-bound" — both still hold. But that is a claim about *how much work* the dense half is, and it says nothing about *where that work sits in the graph* or *what the accelerator costs to feed*. A byte-count argument is not a placement argument.

This is a performance verdict, not a correctness one: GPU output was token-identical to CPU output.

Two things it does **not** settle: a discrete GPU with its own VRAM removes the second cost entirely, so none of this transfers to a desktop; and prefill — batched and compute-bound, the regime a GPU is actually suited to — is unmeasured, since every cell here is decode.

Full method and numbers: `docs/bench-data/2026-07-27-gpu-dense-offload/findings.md`.

### Two real bugs found by running it

- **The streamer segfaulted on the first token with the GPU on.** Pinning the expert *weights* to the CPU was not enough: the streamer also *reads* the routing out of the graph from the host, and the router had been scheduled on the GPU, where `->data` is a device offset — non-NULL, so the existing guard waved it through. Fixed by pinning `ffn_gate_inp` alongside the experts, plus a host-residency test that now fails the session at load with an explanation instead of crashing mid-decode (7de5a53).
- **The app could not load OpenCL at all.** It is not an NDK library, so `<uses-native-library>` is required; without it the process dies before `main`, and the second failure mode it produces when you widen `LD_LIBRARY_PATH` is thoroughly unobvious (f5a32f3). Documented in `docs/gpu-offload.md`.

### Disposition

I'd suggest **not merging the feature as a recommended path**, and treating this as the sidecar (#90) case: the value delivered is the measurement and the mechanism, not the code. Three options, in the order I'd rank them:

1. **Close unmerged, keep the findings.** Cherry-pick the two unrelated fixes (the `printf` `%` bug and the PowerShell build-script robustness) and land `docs/bench-data/2026-07-27-gpu-dense-offload/` on `main` so the negative is on record.
2. **Merge as-is.** Everything is off by default and gated behind an opt-in build, the gates are green, and the toggle then exists to reproduce the result — at the cost of carrying an OpenCL path nobody should turn on.
3. **Keep the branch open** pending the two unsettled questions (desktop discrete GPU, prefill).

Helldez:
## Final measurement: not a lever here — and the two earlier verdicts in this thread were both wrong

Third reading of the same experiment, and the first that survives its own repeat. `Qwen3-30B-A3B`, `--ubatch 256`, cache 2000, five cells with the CPU first **and** last so drift is bounded rather than assumed:

| `--gpu-layers` | tok/s | graph splits | CPU occupancy |
|---|---:|---:|---:|
| 0 (CPU) | 2.941 … 2.778 | 1 | 88 → 90% |
| 48 (all) | 2.768 | 96 | **54%** |
| 12 | 2.866 | 24 | 62% |
| 1 (output head) | 3.029 | 2 | 68% |

Within the pass, throughput orders perfectly and inversely by split count, and **the GPU visibly takes work off the CPU** — occupancy falls from 88% to 54%. Drift-corrected, `--gpu-layers 1` looked like +7%.

**It did not replicate.** Alternating that cell against the CPU gave 2.934 against **3.116** — the opposite ordering by a similar margin. Across both passes the distributions overlap completely: this device's run-to-run noise is **±6%**, larger than every effect measured except the full-offload penalty. Only the negative survives repetition.

Mechanism, which is settled even though the verdict is small: the dense and expert halves **interleave** (dense → experts → dense → experts, ×48), so a two-device split crosses the boundary twice per layer and the boundary tax eats the freed CPU time. `--gpu-layers 1` comes closest only because the output head is the one dense component with no expert layer after it — 2 splits instead of 96 — and one matmul is too small a share of the per-token work to clear the noise.

### Corrections to this thread

1. **"2.6x slower, the GPU adds memory pressure"** — wrong. That was *our* `n_ubatch = n_ctx` reserving 1203 MiB of compute buffer, pushing the system into reclaim. Capping it: 1.207 → 2.752 tok/s, 5 958 → 1.06 major faults/token. Hence `--ubatch N` (28b1e08), which is not a GPU fix — the CPU path reserves 240 avoidable MiB of its own.
2. **"`--gpu-layers 1` wins ~7%"** — wrong. One run per cell cannot see through ±6%.

Both are recorded in `docs/bench-data/2026-07-27-gpu-dense-offload/findings.md` rather than quietly edited away, along with method notes: cap the ubatch, bracket sweeps with the baseline at both ends, and expect the phone to doze and kill detached runs (`scripts/bench-gpu-sweep.sh` writes per cell for that reason).

### What could still change the answer, both untested

- **Contiguous layer blocks.** 96 splits is a property of *this* partition, not of offload as such. Offloading the first N layers **whole** — experts included, resident rather than streamed — puts the boundary in one place. llama.cpp already expresses per-block placement with a regex over `blk.<il>.`, so the CPU pin could be layer-aware instead of global. Cost: ~363 MiB of resident experts per layer that the cache no longer gets.
- **The cache curve, not a single point.** Full offload leaves decode more I/O-bound (compute 0.187 against flash 0.648 s/token), so cache and lane levers should buy *more* with the GPU on. The right comparison is the cache sweep under each backend. Note the GPU frees no RAM — the dense weights move from anon buffers to OpenCL buffers, both the same physical LPDDR.

### Disposition

Unchanged from before, and now better supported: **the value here is the measurement and the mechanism, not the feature.** `--ubatch` is worth keeping on its own merits (it is a real, invisible memory budget on the CPU path too). The GPU path is worth keeping only as the apparatus that makes these numbers reproducible.

Helldez:
## Measured, and the description now says so

This PR opened with a section titled *"Status — not measured"*. It has been measured since, three times, and the description above has been rewritten to lead with the result instead of the argument. Recording here what changed, so the history is not just silently replaced.

**Verdict: full dense offload is −27% and offloading the output head alone is −16 to −21% on `Qwen3.6-35B-A3B`, both outside this device's ±6% noise and consistent across a palindrome.** On the earlier `Qwen3-30B-A3B` the same cells read neutral; that is explained rather than contradicted — the CPU baseline has since got faster (`--drop-cold-experts` is worth ~+18% on this model), so a fixed boundary-crossing cost is a larger share of a smaller total.

Two corrections along the way are worth more than the verdict itself:

1. The first on-device number said **2.6× slower** and was attributed to the GPU sharing LPDDR. That was wrong. It was this engine's own compute-buffer reservation — `n_ubatch` was pinned to `n_ctx`, reserving 1203 MiB with a GPU in the graph — pushing an already-tight system into reclaim at 5 958 major faults per token. Capping it moved the same cell from 1.207 to 2.752 tok/s. `--ubatch N` exists because of this, it applies to the CPU path too, and it is the only part of this PR that is unambiguously worth having.

2. The second pass found `--gpu-layers 1` worth **+7%** after drift correction. A repeat reversed the sign. One run per cell decides nothing at ±6% noise, and the +7% was noise.

**The original description already contained the right answer and did not follow it.** Under "Status — not measured" it listed, as a risk to check first: *"decode is batch-1, i.e. GEMV, which is memory-bound — a GPU's advantage there is bandwidth, not FLOPs, and it shares the same LPDDR."* That is exactly what the measurements found, and it outranks the byte-count argument that motivated the feature. A byte-count argument is not a placement argument — the same shape of error as a hit-rate curve not being a throughput argument.

### Scope of the claim

Decode only, on an integrated mobile GPU. **Prefill is untouched by all of this** and is the regime a GPU actually suits: batched, compute-bound, independently measured at ~2.3× on this device. That experiment is not in this PR and is the one place the GPU still has a plausible case.

### Why it still merges

Nothing here is a correctness problem — output is token-identical at every placement. The flags ship gated and off, with the resolved device reported everywhere so a benchmark cannot silently mistake a CPU run for a GPU one. What merges is a reproducible measurement, `--ubatch N`, and a documented negative result, not a recommendation.

The companion question — whether the routed *experts* can move to the GPU at all — is answered separately and negatively (0.22×) on `feat/gpu-expert-slots`. It needs seams carried on a llama.cpp fork branch and deliberately stays out of this tree.

Helldez:
Closing unmerged. The feature measures negative on the hardware this engine targets, and this project's convention for a measured negative is to close it, document it and tag it — not to merge it gated-off. The same treatment as the expert sidecar (#90) and the per-layer LFU cache.

**What was measured.** Full dense offload −27%, output head alone −16 to −21% on `Qwen3.6-35B-A3B`, both outside this device's ±6% noise and consistent across a palindrome. On the earlier `Qwen3-30B-A3B` the same cells read neutral; that is explained rather than contradicted — the CPU baseline has since got faster, so a fixed boundary-crossing cost is a larger share of a smaller total. Output is token-identical at every placement, so this was never a correctness question.

**Why it cannot be tuned into a win.** The dense and expert halves interleave, so a two-device split crosses the boundary twice per layer — 96 graph splits against 1. The GPU does free the CPU (occupancy 88% → 54%) and the boundary tax eats all of it. Underneath that is the reason no partition escapes: batch-1 decode is bandwidth-bound GEMV, and an integrated GPU shares the CPU's memory bus. The companion experiment (#105) confirmed the same floor from the opposite direction — even a fully resident model with nothing streamed runs at ~0.79×.

**What is being salvaged**, in #106: `--ubatch N` and two unrelated fixes. The reservation coupling that knob breaks is the engine's own — 320 MiB of compute buffer at `n_ctx` 2048, scaling with the context and not with the work — and it applies to every configuration, GPU or not. It was found here only because the GPU made the reservation large enough to notice.

**What stays and stays reachable.** The branch and the tag `gpu-dense-offload-refuted`, so the measurement is reproducible rather than merely asserted, and so the next person to reach for a GPU on a phone finds the numbers before spending the week. The evidence lives in `docs/bench-data/2026-07-27-gpu-dense-offload/` on that branch, including the two wrong verdicts that preceded the right one — a 2.6× "GPU penalty" that was our own compute buffer, and a +7% that did not survive repetition.

**What remains open.** Prefill, which is batched and compute-bound and independently measures ~2.3× on this device. Every number in this PR is decode. That is a different experiment and it is the only GPU question here still worth asking.

PR #105, opened 2026-07-27
research(gpu): stream MoE experts into GPU memory — works, and loses by 4.6x

Helldez:
**Not for merge.** Opened so the experiment, the code and the numbers are on the record; closed immediately because the measurement says the design cannot win on this class of device, and because it depends on two seams carried on a llama.cpp fork branch that `main` does not take.

## The idea

`docs/gpu-offload.md` said the routed experts can never move to a GPU: the streamer rebinds an expert tensor's `->data` to a host buffer it refills every token, and a device buffer has no host address to rebind.

That argument is about rebinding a **pointer**. It says nothing about rebinding the **contents**. Allocate each expert tensor with only `N` experts along dim 2 and it becomes a residency pool: keep an LRU over the slots, read a missing expert from flash with O_DIRECT, write it into a slot, and rewrite the routing so the dispatch names slots instead of experts. The weights never move; what they hold does.

Two properties of the backend make it cheap enough to be worth building:

- the OpenCL convert kernels address the expert dimension as a whole multiple of the per-slice size and **never read `ne02`**, so converting one expert into an arbitrary slot needs nothing but sub-buffer offsets and a depth-one dispatch — no new kernels, no kernel edits;
- the router keeps the full expert count regardless, because `build_moe_ffn` takes it as an argument rather than reading it off the tensor.

## It works

Bit-exact slice writes, coherent text, on a 20.8 GB MoE with 256 experts held 24 at a time. Flash → GPU slot → id remap → `MUL_MAT_ID`, proven on real hardware.

## It loses, by 4.6×

**1.026 tok/s against 4.689** for the CPU streaming path. Same binary, same model, same prompt, and critically **the same bytes off the flash at the same hit rate** — 236.7 vs 235.7 MiB/token, 56.5% vs 54.4%. The CPU cell ran *second*, on a hotter device, and still won.

The token decomposes as compute 0.244 + flash read 0.302 + slice upload 0.406 + id readback 0.023 seconds. Two results are worth having independently of the verdict:

**The per-layer synchronisation is not the problem.** Draining the GPU queue once per MoE layer so the host can read the routing costs **2.4% of the token**. The published prior art that killed this design on another backend measured ~400 µs per sync and concluded it was fatal. Here it is noise. That objection is refuted, and it is the interesting half of this experiment.

**The upload is the problem, and fixing it would not be enough.** It runs at 0.57 GiB/s against 4.8–5.9 GiB/s of raw device bandwidth, because every projection of every miss pays a blocking write plus a blocking convert dispatch — roughly 400 driver round-trips per token, each at full latency. A second command queue, double-buffered staging and a non-blocking wait would recover most of it. But set upload **and** read to zero — a perfect pool that never misses and never waits — and the floor is still **0.267 s/token, i.e. 3.7 tok/s**, below what the CPU path already delivers. The idealised version loses to the shipping version.

Underneath is the compute: **0.244 s/token against the CPU's 0.126**, uniform across all 40 layers (23.3 ms vs 4.77 ms mean per layer, no pathological layer, nothing to fix). Batch-1 decode is matrix-vector work bound by memory bandwidth, and an integrated GPU reads the same LPDDR as the CPU while adding per-dispatch overhead and a weight format it cannot read natively. A GPU is a batch engine; token-by-token generation is the one workload that never batches.

## Sizing, the one large lever inside the design

`N` must clear one token's working set — `k` experts per layer — or the pool hits **exactly 0%**, the same cliff the CPU cache has and for the same reason. `N=8` on a k=8 model is therefore the worst possible choice; `N=8` → `N=24` is **+55%** (0.720 → 1.116 tok/s). One slot costs ~70 MiB across all layers on this model, so the ~4.5 GiB the driver reports leaves room for roughly 35 once the dense half is resident. This is a real lever and it is nowhere near enough to close a 4.6× gap.

## The complete map

Every GPU placement measured against the same CPU-only baseline on this device:

| placement | relative decode throughput |
|---|---:|
| routed experts streamed into GPU slots (this PR) | 0.22× |
| full dense offload, 40 layers (#104) | 0.73× |
| output head only, 1 layer (#104) | 0.80 – 0.84× |
| whole model resident, nothing streamed | ~0.79× |
| **CPU only** | **1.00×** |

The fourth row is what closes the question: that is the GPU's best possible case — everything resident, nothing streamed, no host in the loop at all — and it still loses. No placement strategy can beat a configuration that has already removed every cost a placement strategy could remove.

## What it depends on

Two seams on `Helldez/llama.cpp`, branch `bmoe/opencl-unified-memory`, both inert unless used:

- `llama_model_params::moe_expert_slots` — the loader allocates routed-expert tensors narrowed to `N` along dim 2 and skips loading their data. The gguf shape check still runs first and validates against the file, so this narrows an already-validated tensor rather than accepting a mismatch.
- `ggml_backend_opencl_set_expert_slice()` and `ggml_backend_opencl_set_moe_slot_hook()` — per-slice upload, and a substitution hook at the top of `ggml_cl_mul_mat_id`. Substitution rather than in-place id rewriting because the routed ids are also consumed downstream by the top-k weighting, and the pool must live in a buffer outside the graph allocator.

The engine side compiles without them: CMake feature-detects the hook and builds a stub that refuses the flag, so a stock tree is unaffected. That is why the code could sit on this branch at all — but the flag is useless without the fork, and carrying a fork in `main` is the one thing this project does not do.

## What stays open

**Prefill.** Batched, compute-bound, the regime a GPU actually suits, and independently measured at ~2.3× on this device. Every number here is decode, and the prefill cells in the data are 7-token prompts — far too short to say anything.

## Where the evidence is

`docs/bench-data/2026-07-27-gpu-expert-slots/` — findings, raw per-token CSVs for both cells, the per-layer breakdown, and the console summaries. Method notes included, in particular: instrument the new path *before* comparing it. In slot mode the streamer's own counters read zero, so the first version of this comparison had a 0.73 s/token hole in it that looked like compute.

Revisit if a backend ever grows a slot-allocation API upstream, or on hardware whose GPU does not share the CPU's memory bus. Until then the measurement stands.


Helldez:
Closing unmerged, as intended when it was opened.

The design is sound and proven — flash to GPU slot to remapped routing to `MUL_MAT_ID`, bit-exact — and it is 4.6x slower than streaming the same experts on the CPU, with a compute floor underneath that no amount of upload engineering reaches. `main` also does not carry a llama.cpp fork, and this needs two seams on one.

What this experiment leaves behind, in order of durability:

1. **The per-layer host synchronisation is cheap** (2.4% of the token), which refutes the objection that killed this design elsewhere. That result outlives the verdict.
2. **The map is complete**: every GPU placement loses on this class of device, including a fully resident model with nothing streamed. Decode is bandwidth-bound GEMV and an integrated GPU shares the CPU's memory bus.
3. **The branch and the fork stay published** so the measurement is reproducible rather than merely asserted.

Reopen if a backend grows a slot-allocation API upstream, or on hardware whose GPU has its own memory. The remaining GPU question worth asking on this device is prefill, which is a different measurement.

Helldez:
Cross-reference: the sibling PR #104 (dense-path offload) is now also closed unmerged, on the same grounds and with the same tag treatment. Between the two, every GPU placement on this class of device is measured and documented:

| placement | relative decode throughput |
|---|---:|
| routed experts streamed into GPU slots (#105) | 0.22x |
| full dense offload (#104) | 0.73x |
| output head only (#104) | 0.80 - 0.84x |
| whole model resident, nothing streamed | ~0.79x |
| **CPU only** | **1.00x** |

`--ubatch N`, the one lever from that line of work that pays and has nothing to do with the GPU, is salvaged to `main` in #106.

PR #106, opened 2026-07-27
feat(engine): --ubatch N — the compute-buffer reservation nobody was watching

Helldez:
Salvaged from #104, which is being closed unmerged because the feature it was built for measures negative. This is the part of that work that has nothing to do with the GPU and should not die with it.

## The problem

Compute buffers are reserved for the worst-case graph, and this engine had always set `n_ubatch = n_ctx` so that any fitting prompt prefills in one pass. That quietly ties **resident memory** to the context rather than to the work:

| n_ctx | compute buffer reserved |
|---|---:|
| 2048 | 320 MiB |
| 512 | 80 MiB |

Exactly 4× for a 4× context. On an engine whose entire problem is that the expert cache and the dense weights compete for the same RAM, that is a real budget — and it was invisible, because nothing reported it and nothing could change it.

## The knob

`--ubatch N` sets the widest graph computed at once. `0` (the default) keeps the previous behaviour exactly, so this is inert unless asked for.

**Decode is unaffected** — a decode graph is one token wide whatever this says. The cost is prefill throughput: a long prompt is processed in more, smaller passes. That is the trade, and it is worth having explicitly on a device where the alternative is the kernel reclaiming the weights out from under the run.

`validate()` rejects a negative value, and one above `n_ctx`: reserving for a batch that cannot arrive is the inverse of what the knob is for.

## How it was found

Chasing what looked like a catastrophic slowdown — nearly 6 000 major faults per token — which turned out not to be the workload at all. It was this reservation pushing an already-tight system into reclaim. Capping it moved the affected configuration from 1.207 to 2.752 tok/s, and the fault count from 5 958 to 1.06 per token.

That measurement happened during GPU work, where the reservation was largest, but the coupling it exposed is the engine's own and applies to every configuration.

## Also fixed

Two unrelated things found alongside it:

- A stray `%` in the `--dense-weights` help text was read by `printf` as a conversion specifier, so that line printed garbage and read past the argument list.
- `scripts/build-android.ps1` judged `cmake` by whether it wrote to stderr rather than by its exit code, which broke it under Windows PowerShell 5.1 — the NDK toolchain prints progress to stderr, and 5.1 turns that into a terminating error under `ErrorActionPreference = Stop`. It now checks `$LASTEXITCODE`, so the script behaves the same in both shells.

## Verification

- Host byte-identity gates **7/7** green.
- `config_validate` extended with the new bounds.
- clang-format 18 clean.
- No submodule bump: this branch sits on `main`'s llama.cpp pointer.

## What is deliberately not here

Everything GPU. `--gpu`, `--gpu-layers`, `--gpu-require`, the OpenCL build path and the manifest change stay on `feat/gpu-dense-offload` (#104, closed) with the measurements that refuted them; expert residency in device memory stays on `feat/gpu-expert-slots` (#105, closed). Both are tagged and reachable, and neither belongs in `main`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 /out/
 cmake-build-*/
 
+# OpenCL headers + ICD loader, fetched by scripts/fetch-opencl-android.ps1 for the optional
+# GPU build. Third-party sources we do not vendor: a build prerequisite, not our code.
+/third_party/opencl-sdk/
+
 # Models and large artifacts (never commit weights)
 *.gguf
 *.bin
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,30 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [Unreleased]
+
+### Added
+- **`--ubatch N`: the widest graph computed at once, decoupled from the context.** Compute buffers
+  are reserved for the worst-case graph, and the engine had always set `n_ubatch = n_ctx` so that
+  any fitting prompt prefills in one pass — which quietly ties **resident memory** to the context
+  rather than to the work. Measured at n_ctx 2048: **320 MiB of compute buffer, falling to 80 MiB
+  at 512**, scaling exactly with the context. On an engine whose whole problem is that the expert
+  cache and the dense weights compete for RAM, that is a real budget, and it was invisible.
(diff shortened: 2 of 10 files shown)


Helldez:
Added a second commit: `.gitignore` now covers `third_party/opencl-sdk/`. `scripts/fetch-opencl-android.ps1` drops 5.3 MB of Khronos headers and an ICD loader there, nothing was ignoring them, and they sit untracked in any tree that ever configured a GPU build — one careless `git add -A` from landing in a public repo. The GPU build is not in `main` and is not coming, but the directory outlives the experiment in working trees that tried it.

PR #135, opened 2026-07-29
feat(dense): --dense-embd-mmap, keep the embedding table out of the pinned set

Helldez:
## What this is

The dense-weights policy applies **one mode to every non-expert byte**. That quietly treats the two
largest dense tensors as equals, and a compute trace says they are nothing alike.

Measured with `--compute-trace` on a host, per decoded token:

| tensor | Qwen3-30B | bytes read per token | node cost |
|---|---|---|---|
| `output.weight` (head) | 243 MiB Q6_K | **all of it** — a matmul over the whole vocabulary | 8.94 ms |
| `token_embd.weight` | 167 MiB Q4_K | **one row, 1152 bytes** — a `GET_ROWS` lookup | 0.26 ms |

The head must stay resident: evicted, every token would refault the entire tensor from flash. The
table need not. Pinned, it holds 167 MiB — 273 on Qwen3.6, 380 on gpt-oss-120b — for about a
kilobyte of use per token, memory the expert cache could hold instead.

`--dense-embd-mmap` (off by default) holds the embedding table back from the anon/`ahwb` set and
leaves it mmap'd, so the kernel keeps only the 4 KiB pages a conversation actually touches. That
converges to the distinct token ids in use; the cost is one fault the first time an id appears, not
one per token.

Verified on Qwen3-30B: the anon set drops from **951 MiB to 784 MiB**, 435 buffers to 434,
generation unchanged.

```
bmoe: --dense-embd-mmap — token_embd (166 MiB) left mmap'd
bmoe: dense-weights=anon — 784 MiB in 434 anon buffers
```

## Why it is off by default

The win is **memory handed back to the expert cache**, not time. It converts to tok/s only in a
regime where more cache buys hits — below the `token_demand` cliff it buys nothing. The mechanism
ships now; the flag flips after the A/B on device, which this PR does not claim to have run (no
device was attached while writing it).

## Tied models

Refused, with a message, when the gguf has no separate `output.weight`: there the table *is* the
head, read whole every token, and the argument inverts. The check is a runtime lookup of llama.cpp's
canonical tensor names, not an architecture list. **Untested** — no tied MoE gguf was available
locally; the untied path is what the verification above exercises.

## What else the trace settled

Recorded in `docs/android-memory.md` because it closes two questions rather than opening them:

- **The head is bandwidth-bound and has no software lever.** Halving the threads (8 → 4) leaves the
  `result_output` node at 8.94 → 8.97 ms, and two different head sizes run at the same 26.6 /
  27.5 GB/s. Its cost is exactly `bytes / bandwidth`, so no kernel, threading or coalescing change
  can move it — only fewer bytes, i.e. a lower `--output-tensor-type` at quantization time. On
  gpt-oss-120b the head is 587 MiB Q8_0, where that would be worth the most.
- **The final norm is not a cost.** All 49 `RMS_NORM` nodes together are 1.0–1.4 ms/token.

## Gates

`ctest` 7/7 pass. The change touches which tensors the dense policy adopts, not the streamer or the
seam, and the streamed-equals-resident gates are unaffected — but they were run before and after
`clang-format` anyway.

## Not in this PR

No Android settings toggle yet: the app should expose this once the device A/B says which way it
goes, so the setting arrives with a measured default rather than a guess.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,28 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [Unreleased]
+
+### Added
+- **`--dense-embd-mmap` — keep the embedding table out of the pinned dense set (off by default).**
+  The dense policy applied one mode to every non-expert byte, which treated the two vocabulary
+  tensors as equals. A compute trace says they are not: on Qwen3-30B the output head is 243 MiB and
+  **all of it** is read every token (a matmul over the whole vocabulary, 8.94 ms/token), while
+  `token_embd` is 167 MiB of which **one row, 1152 bytes**, is read (a `GET_ROWS`, 0.26 ms). Pinning
+  both spends 167 MiB — 273 on Qwen3.6, 380 on gpt-oss-120b — for about a kilobyte of use per token.
+  With the flag on, the table stays mmap'd and the kernel keeps only the pages a conversation
+  touches; the anon set drops from 951 to 784 MiB on Qwen3-30B. Refused, with a message, on a model
+  that ties the table to the head, where the table is the head. The win is memory handed back to the
+  expert cache, so it converts to tok/s only where more cache buys hits — the flag stays off until
+  that A/B runs on device. `dense_embd_mmap` joins the CSV config line.
+
+### Documented
+- **The head is bandwidth-bound; the final norm is not a cost.** Halving the threads leaves the
+  `result_output` node at 8.94 → 8.97 ms and two head sizes run at the same 26.6 / 27.5 GB/s, so its
+  time is exactly `bytes / bandwidth` and no kernel or threading change can move it — only fewer
+  bytes, i.e. a lower `--output-tensor-type` at quantization time. All 49 `RMS_NORM` nodes together
+  are 1.4 ms. `docs/android-memory.md` gains the measurements.
+
 ## [0.18.0] - 2026-07-28
 
 ### Added
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -399,6 +399,11 @@ static void print_usage(const char * argv0) {
         "                          ahwb = as anon, but into dma-buf memory the kernel may not reclaim\n"
         "                          at all — not even to zram, which is what anon still pays for.\n"
         "                          Android-only; measured +17.9%% on a long generation, off by default)\n"
+        "      --dense-embd-mmap   keep the embedding table OUT of the anon/ahwb dense set, mmap'd.\n"
+        "                          It is read one row (~1 KiB) per token but is as large as the\n"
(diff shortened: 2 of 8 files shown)


Helldez:
Measured on device, and the answer is no. Closing.

One cell per arm, Qwen3-30B-A3B-Q4_K_M, shipping recipe
(`--moe-stream --cache-mb auto --io-threads 4 --overlap --dense-weights anon --drop-cold-experts 0.75`),
64 tokens, the flag as the only variable.

| | A (baseline) | B (`--dense-embd-mmap`) |
|---|---|---|
| **decode** | **6.248 tok/s** | **5.232 tok/s** |
| median wall/token | 151.3 ms | 159.6 ms |
| median compute/token | 96.7 ms | 102.5 ms |
| model load | 29.1 s | 25.0 s |
| prefill | 6.3 s | 10.3 s |
| TTFT | 35.42 s | 35.32 s |
| cache budget | 4158 MiB | 4360 MiB |
| cache hit | 75.9% | 76.9% |
| dense anon set | 951 MiB | 784 MiB |
| `dense_resident_frac` | 0.82 | 1.00 |

Every mechanical prediction in the PR held: 167 MiB left the anon set, the cache took 202 MiB more,
the hit rate rose a point, and the smaller anon set stayed fully resident instead of at 0.82. In
steady-state decode the access-pattern argument held too — median major faults per token are **0
under the flag against 1 without it**, so the table genuinely is not refaulting.

It buys nothing, and it costs.

**The load is displaced, not avoided.** The flag saves 4.1 s of model load by not reading those
167 MiB via O_DIRECT, and repays it almost exactly in prefill, 6.3 → 10.3 s. TTFT comes out even to
a tenth of a second. Demand-faulting those pages 4 KiB at a time costs what reading them once
sequentially costs — which is the demand-fault storm already documented in `docs/android-memory.md`,
applying to this change as much as to the case it was written for.

**The freed memory does not convert.** 202 MiB more cache bought one point of hit rate and no
throughput; compute rose 96.7 → 102.5 ms. Same shape as the layer-LFU result: hit rate is not
throughput.

Caveats stated plainly: one cell per arm, no cooldown between them, B ran second (battery 31.0 →
31.3 °C). Mean and median disagree — −16% against −5.2% — so the tail is noisy and this is a clear
negative signal rather than a settled verdict. Repeated order-swapped cells could tighten it, but
the better of the two readings is still a loss and TTFT is a wash, so there is nothing here worth
buying with more device time.

What the investigation did settle is kept out of this branch and stands on its own: the output head
is bandwidth-bound (halving threads leaves the `result_output` node at 8.94 → 8.97 ms; two head
sizes run at 26.6 / 27.5 GB/s), so only a lower `--output-tensor-type` at quantization time can move
it, and the final norm is not a cost at all (all 49 `RMS_NORM` nodes together are 1.4 ms/token).
Those belong in the docs on their own, not behind a refuted flag.

PR #139, opened 2026-07-30
research(moe): price a row-sparse expert matmul, and refute it

Helldez:
**Draft on purpose — not for merge.** This prices the "LLM in a Flash" neuron-selection idea,
adapted to a MoE engine, and the answer is that it does not pay here. Opening it so the
measurement and the reason are on the record rather than in a chat log.

## The question

Top-k routing already skips whole experts. The open question was whether, *inside* a routed
expert, a small share of the intermediate neurons carries the output — in which case a
row-sparse expert matmul would cut the bytes read and the arithmetic together.

## What was built

Two knobs on the engine and a kernel on the fork, all off by default, all measurements
rather than features:

| flag | what it does | what it prices |
|---|---|---|
| `--gate-sparsity` | observes how concentrated each routed expert's intermediate vector is | is there sparsity at all |
| `--expert-row-sparsity F` | zeroes the lowest-magnitude `F` of that vector | **quality** — costs nothing, saves nothing |
| `--expert-row-stride N` | evaluates only every `N`th output row of the up-projection | **speed** — reply is meaningless by construction |

The last one needs `ggml_cpu_set_expert_row_stride` from
[`Helldez/llama.cpp@bmoe/row-sparse-experiment`](https://github.com/Helldez/llama.cpp/tree/bmoe/row-sparse-experiment),
one commit on top of the existing expert-ready-hook branch. No in-tree patch, per the
project's first hard rule; the fork branch is the sanctioned fallback and was agreed before
it was written.

They are deliberately three knobs and not one: quality and speed cannot be read off the same
cell, and merging them would produce a number nobody could interpret.

## The measurements

**Sparsity exists, but it is not structure.** `--gate-sparsity` over two models:

| | rows for 95% of L1 mass | mass at 25% of rows | at 12.5% | exact zeros |
|---|---|---|---|---|
| Qwen3.6-35B-A3B (512-wide experts) | 65.2% | 64.6% | 44.1% | 1.14% |
| Qwen3-30B-A3B (768-wide experts) | 65.6% | 65.0% | 44.9% | 0.00% |
| swiglu of gaussian noise | 61.9% | 71.6% | 53.0% | — |
| plain gaussian | 75.0% | 51.6% | 30.8% | — |

Read the third row first: a SwiGLU of *independent gaussian noise* is more concentrated than
either real model. The concentration these models show is what the nonlinearity does to an
unstructured vector, and slightly less of it.

**Quality is not the blocker, though.** A mass-retention criterion turned out to be far too
conservative — the literature (CATS at 50%, TEAL at 40-50%, and an MoE intra-expert result at
up to 90%) says these rows can go for 1-2% on downstream tasks, and a smoke test held the
exact answer with 75% of rows zeroed. The blocker is on the other side.

**The ceiling, measured rather than deduced.** Prefill of a fixed 588-token prompt, so every
cell does identical work; cache pinned at 3000 MiB; baseline run first *and* last:

| stride | rows of up computed | compute | vs baseline |
|---|---|---|---|
| 1 | 100% | 20.49 s | +0.6% |
| 2 | 50% | 19.62 s | −3.6% |
| 4 | 25% | 18.74 s | −8.0% |
| 8 | 12.5% | 18.62 s | −8.5% |
| **0** | **0%** | **18.37 s** | **−9.8%** |
| 1 | 100% | 20.23 s | −0.6% |

Deleting the up-projection outright costs **9.8% of prefill compute**. No row policy can beat
cancelling every row, so that is the bound. On the reference device (compute ~77% of wall) a
CATS-style policy is worth ~+2.9% tok/s at 50% sparsity and ~+6.5% at 75%.

For comparison, both already shipped and neither needs a kernel: `--n-expert-used 6` measured
+22-24%, `--drop-cold-experts` +55-84%.

## Why the down-projection cannot be included

`ffn_up_exps` is `{n_embd, n_ff, n_expert}`: a neuron is a whole output row spanning an integer
number of quantization blocks, so skipping it is clean. `ffn_down_exps` is
`{n_ff, n_embd, n_expert}`, where those same neurons are scattered positions *inside* every
block along the reduction axis — nothing is skippable there without dequantizing the block
regardless. CATS and TEAL do not meet this because they run fp16 on GPU, where a gather along
K is cheap and quantization blocks do not exist. On a quantized gguf the symmetry breaks.

The I/O half fails for a related reason: the reads are issued at the routing node, before the
gate that would decide the mask exists, and a magnitude mask selects *scattered* ~1.1 KB rows —
the sub-256 KB regime `tools/bmoe-iobench` already measured as bandwidth death, and the
mechanism that closed the sidecar (#90).

## Two traps this measurement fell into, recorded so the next one does not

1. **A tok/s that was not real.** The first decode sweep showed +70%. It was an artefact: with
   the numerics broken the model degenerates into a `<think></think>` loop, which routes to a
   tiny repeating expert set, lifts the cache hit rate to 90.7% and collapses flash reads. The
   kernel runs *after* the reads and cannot touch them — when a number improves where a change
   cannot reach, the experiment is broken, not the change. Only the compute term is readable.

2. **`--compute-trace` overstates small nodes.** It budgeted the up-projection at 19.1% of
   decode compute; removing the work entirely says ~10%. A barrier per node inflates small
   nodes by roughly 2x. **Any attribution in this repo taken from a per-op trace deserves
   re-reading against a measurement that removes work and watches what disappears.** That is
   probably worth more than the feature this PR set out to build.

## Validation

- Host gates 7/7 green on this branch (the diagnostics are off by default, so the byte-identity
  path is untouched).
- `--expert-row-sparsity` zeroes what it is asked to: 1,965,696 of 2,621,440 rows at F=0.75.
- clang-format 18 clean.
- Android app compiles; the two settings ship off.

## Disposition

Closing unmerged. The engine keeps nothing; the fork branch stays as the record of what the
kernel looked like. What is worth carrying forward is a docs entry — the layout argument, the
9.8% ceiling, and the `--compute-trace` correction — which belongs in `docs/roadmap.md`
alongside the other measured negatives, in its own PR.

### DIFF
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -423,6 +423,10 @@ static void print_usage(const char * argv0) {
         "                          predicted below the drop threshold are not speculated.\n"
         "      --predict-spec-max N  speculated predicted misses per layer [0..8] (default 2;\n"
         "                          0 = retention only, the prediction spends no flash at all)\n"
+        "      --gate-sparsity     diagnostics: measure how concentrated the vector each routed\n"
+        "                          expert's down projection consumes is — what a row-sparse expert\n"
+        "                          matmul could skip. Observes only; costs a barrier per layer and\n"
+        "                          a sort per routed slot, on every token including the prompt.\n"
         "      --list-archs        print supported MoE architectures and exit\n"
         "\n"
         "  Env overrides (flag wins): BMOE_CACHE_MB, BMOE_IO_THREADS, BMOE_PROGRESS, BMOE_OVERLAP, BMOE_PREFETCH, "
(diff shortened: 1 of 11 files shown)


Helldez:
Closing unmerged: the measurement is the deliverable and it is negative. Deleting the up-projection outright costs 9.8% of compute, so no row policy can return more than that, and the down-projection is not row-skippable on a quantized gguf at all. The kernel stays on `Helldez/llama.cpp@bmoe/row-sparse-experiment` and this branch is preserved as the tag `row-sparsity-refuted`; the branch itself is deleted. The two findings worth carrying forward - the weight-layout argument and the `--compute-trace` inflation of small nodes - go into `docs/roadmap.md` in their own PR.

PR #143, opened 2026-08-01
perf(io): read expert slices straight into the cache, with no bounce copy

Helldez:
## What

`--odirect-zero-copy` (off by default): expert slices are read **straight into the cache buffers**, with no bounce buffer and no copy. Measured on device: **+20% tok/s**, identical bytes read, byte-identical output.

This is not a decode feature — every expert read in the engine goes through this path, so it applies to the plain streaming baseline, expert dropping, the prefetchers, everything.

## Why the copy existed

O_DIRECT requires the file offset, the transfer length and the buffer address to be aligned. Expert slices are a whole number of pages long, but a gguf's tensor data does not begin on a page boundary — `general.alignment` is 32 by default, and on the model family measured here **every** expert offset lands a constant **1152 bytes** past one:

```
offset 898471040  mod 4096 = 1152
offset 900830336  mod 4096 = 1152
offset 903779456  mod 4096 = 1152      (every read in a 12k-read trace)
```

So the reader could not ask the kernel for the payload directly. It pulled the enclosing **aligned window** — which starts 1152 bytes early — into a per-lane bounce buffer, then `memcpy`'d the payload back by that remainder so it landed where `mul_mat_id` looks for it.

The copy was therefore never a safety measure. It was a **shift correction**, costing **200+ MiB a token** (~450 MB/token of DRAM traffic once the write side is counted) on a device whose decode is already memory-bound.

## Why it can go

The shift is only needed because the destination does *not* share the file's remainder — and it can. This engine reserves the per-layer buffers itself (`vm_reserve`, then `tensor->data` is rebound), so it controls their addresses. With the flag on, each buffer is placed at an address carrying its own tensor's remainder. Every expert inherits it, because the per-expert stride is a multiple of the page size. The aligned window then maps onto the buffer at the same relative position and the read goes straight in.

### The overhang, and why it is safe by construction

A read now writes up to `shift` bytes before its slice and the rest of a page after it — bytes belonging to the neighbouring experts. Three facts make that sound rather than lucky:

1. Under this placement buffer and file differ by a **constant** offset, so the overhung bytes receive **their own correct file contents**. Nothing is corrupted even in principle.
2. The written range is **exactly** the range the entry's page commit already covers (`round_down(dst)` … `round_up(dst+slice)`), so no uncommitted page is ever touched. This falls out of the arithmetic, not from a new invariant.
3. Eviction has always released only pages **fully contained** in a slice, never one shared with a neighbour — so a boundary page cannot be decommitted under a lane that is writing it.

A compute thread reading a neighbour's boundary concurrently sees identical values.

## Results (device, Qwen3.6-35B-A3B Q4_0, cache 2000 MiB, overlap, 4 lanes)

Interleaved cells, so thermal drift hits both configurations equally:

| cell | tok/s | compute | stall | read |
| --- | --- | --- | --- | --- |
| off (pass a) | 3.90 | 0.161 | 0.067 | 230.77 MiB/tok |
| **zero-copy (pass a)** | **4.43** | 0.142 | 0.055 | 230.77 MiB/tok |
| off (pass b) | 3.40 | 0.200 | 0.062 | 230.77 MiB/tok |
| **zero-copy (pass b)** | **4.37** | 0.144 | 0.054 | 230.77 MiB/tok |

**+20% median** (3.65 → 4.40), with the win landing exactly where the mechanism predicts: the compute residual falls **0.180 → 0.143 s/token**. Bytes read are identical to the byte in every cell, and the generated text is **byte-identical** with the flag on and off (greedy, same prompt, same md5) — which is the correctness proof.

The phone was warm and its own baseline drifted 3.90 → 3.40 across the run, so the **ratio** is the claim, not the absolutes. Note the last and hottest cell (zero-copy, pass b) still beats the first and coolest baseline cell by 12%.

## On the host gates

The gates **cannot** prove this path, and the PR does not pretend otherwise: on a tiny test model the per-expert stride is not a multiple of the page size, so the placement declines and G8 / G4e demonstrate only that the flag is harmless. Rather than let a silent no-op read as a pass, the engine states which happened:

```
bmoe: odirect-zero-copy ON — 120/120 layer buffers placed to skip the bounce copy   (device, real model)
bmoe: odirect-zero-copy INERT — 0/12 layer buffers placed to skip the bounce copy   (host, tiny model)
```

The device run above is the proof. Gates 7/7.

## Fallbacks

Back to the bounce whenever the remainder would leave the tensor under-aligned for the compute kernels (below 64 bytes), the per-expert stride is not a multiple of the alignment, or O_DIRECT was refused at open. Buffered mode already read straight into the destination and is untouched.

## Scope

Off by default pending a confirmation run on a cool device. Docs: `docs/moe-streaming.md`, CHANGELOG under Unreleased. New CSV preamble key `odirect_zero_copy=`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,7 +45,14 @@ Semantic Versioning.
   (Qwen3.6-35B-A3B Q4_0, cache 2000 MiB, overlap, interleaved cells so thermal drift hits both):
   **+20% tok/s**, 3.65 → 4.40 median, with the compute residual down 0.180 → 0.143 s/token, the
   same bytes read (230.77 MiB/token in every cell) and the generated text **byte-identical** with
-  the flag on and off. This is not specific to any decode feature: every expert read goes through
+  the flag on and off. That figure comes from short single-turn CLI runs, and it hid a leak: off
+  the page boundary the first and last page of every slice is shared with the neighbouring expert,
+  and eviction — which may never free a page a neighbour still uses — left two pages behind each
+  time. Over a long app session that grew the anonymous footprint tens of MiB a turn, went to zram,
+  and came back as major faults (1-7 per token without the flag; 66 then 332 with it, rising turn
+  over turn), making the flag a net LOSS in the app: 4.41-4.57 tok/s against 5.02-5.58. Eviction now
+  releases a boundary page once the neighbour sharing it is gone, and **both numbers are owed a
+  re-measurement on a long session** before either is quotable. This is not specific to any decode feature: every expert read goes through
   this path. The host gates cannot prove it — a tiny test model's per-expert stride is not a
   multiple of the page size, so the placement declines and the gates show only that the flag is
   harmless — so the engine reports which happened (`odirect-zero-copy ON|INERT — n/m layer buffers
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,32 @@ Semantic Versioning.
   (compressed sparse attention, the lightning indexer and its dedicated KV cache) is dense-side
   llama.cpp code, invisible to the streaming seam. Requires a llama.cpp with DeepSeek V4 support
   (the pinned submodule has it).
+- **`--odirect-zero-copy` — read expert slices straight into the cache, with no bounce copy
+  (off by default).** O_DIRECT wants the file offset, the length and the buffer address aligned.
+  Expert slices are a whole number of pages long, but a gguf's tensor data does not start on a page
+  boundary — `general.alignment` is 32 by default, so on the measured model family every expert
+  offset sits a constant 1152 bytes past one. The reader therefore pulled the enclosing aligned
+  window into a per-lane bounce buffer and `memcpy`'d the payload back by that remainder: over
+  200 MiB a token of pure shift correction, on a device whose decode is already memory-bound. The
+  shift exists only because the destination did not share the file's remainder — and it can, since
+  this engine reserves the per-layer buffers itself. With the flag on, each buffer is placed at an
+  address carrying its own tensor's remainder; every expert inherits it (the per-expert stride is a
+  multiple of the page size), so the aligned window maps onto the buffer in place and the read
+  needs no copy at all. The window overhangs its neighbours' slices, which is safe by construction
+  rather than by luck: buffer and file differ by a constant offset under this placement, so the
+  overhung bytes get their own correct file contents, into pages the entry's commit already covers
+  exactly (eviction never releases a page shared with a neighbour). Measured on device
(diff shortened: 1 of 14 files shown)


Helldez:
## Correction: the +20% above was measured where the bug could not show

Running this in the demo app — long session, several turns — it measured **slower** than the baseline, not faster:

| | tok/s | major faults / token |
| --- | --- | --- |
| flag off | 5.02 / 5.58 | **1.2 / 6.7** |
| flag on | 4.57 / 4.41 | **66 / 332** |

Reversing the order of the runs changed nothing, so it is neither warm-up nor thermal drift. The fault count **rises turn over turn** (66 → 332), which is the shape of a leak rather than of a fixed cost, and the process's swap grew 274 → 364 MiB across the session while the flag-off control stayed flat at 272.

### Cause

The placement moves each layer buffer off the page boundary on purpose. That makes the first and last page of **every** expert slice shared with the neighbouring expert — where the aligned placement had no partial pages at all.

Eviction releases only the pages an entry *fully* covers, and it must: it may never free a page a neighbour is still using. Under the aligned placement that rule was exact, because "fully covered" meant all of them. Under the shift it leaks two pages per eviction — bytes the cache counts as freed that the kernel still holds.

The arithmetic matches the measurement: commit covers 145 pages per slice, release covers 143, and a turn evicts ~5200 entries → ~41 MiB a turn, against a measured swap growth of 9 MiB in the first turn and 44 in the second.

### Fix

Eviction now also takes a boundary page once the neighbour sharing it is gone — not resident (`cvalid_`) and with no speculative read outstanding (`spec_remaining_`). Both are read without the lanes' mutex, which is safe in the only direction that matters: staging happens on the eval thread, so nothing raises them concurrently, and a lane can only lower `spec_remaining_` toward zero — a stale read is stale-high and merely defers a release to the next eviction. The outermost pages of a buffer border its own padding, which nobody else owns.

### Why the original measurement missed it

Every device run behind the +20% was **short, single-turn, in a fresh process** (24-40 tokens). A leak has no time to accumulate there. Only the app's long session shows it — the trap this project had already recorded once, that a single-shot bench cannot see a reclaim problem.

**Neither number in this PR is settled.** The +20% and the regression were both measured against a leaking build; the headline needs re-taking on a long app session before it means anything. The flag stays off by default.

Helldez:
Parked, not abandoned. The Windows corruption is tracked in #149 and this branch now carries a guard that falls back to the bounce buffer under \--overlap\ on Windows instead of producing wrong bytes silently, so nothing here is unsafe to build. It stays out of v0.19.0: the flag is experimental, the win it claims is owed a re-measurement after the boundary-page fix, and the gate that should protect it (G4e) is still inert on the tiny fixture. Landing it would mean shipping a feature whose only automated coverage proves nothing.

PR #159, opened 2026-08-03
Use Markov Chain for temporal prediction

supastishn:
This uses a Markov chain to assist with prediction of experts which should achieve higher accuracy.


Helldez:
Sorry for the wait. Before I can judge this I need a number, not a description:

- Run `--predict-log` on a real model with and without the chain and post both. It reports predictor accuracy per layer, it exists for exactly this.
- clang-format 18 on the touched files (CI checks it), and please keep the comments you removed in `session.cpp`.

One thing to know going in: I re-measured prefetch on the phone today and it loses even when the prediction is right (+44% bytes for -14% stall, tok/s down). Better prediction helps only if it costs fewer bytes.

Issue #147, opened 2026-08-01
incredible!


M4RCORENE:
<img width="1216" height="2688" alt="Image" src="https://github.com/user-attachments/assets/e0a76598-1cfe-42b8-84b8-3997bf61a6e4" />
<img width="1216" height="2688" alt="Image" src="https://github.com/user-attachments/assets/3dc2a6e4-bcb2-41db-badb-44f8e40193a5" />

Red Magic 11 Pro

The next model I want to test is the Qwen3.5 397B model 🧐👍

Helldez:
Incredible!!! Can I repost this, maybe on Reddit?
I'm trying to get DeepSeek V4 Flash 0731 running too (look at PR) - and I'm almost there - but I only have a mobile device with 12GB of RAM!

M4RCORENE:
Yes, of course you can post that on Reddit 😊👍

Helldez:
Just a heads-up: you aren't utilizing the full potential because you aren't using O-Direct... you should set up the model in the app, and it should run at double the speed. Let me know.

<img width="937" height="186" alt="Image" src="https://github.com/user-attachments/assets/bf2c7291-a5c5-40bf-b081-474096cb7007" />

M4RCORENE:
Thanks for the tip 😁👍

M4RCORENE:
I already have Direct I/O enabled in the settings, but it says 'I/O buffered (O_DIRECT unsupported on this storage)' in my metrics. What exactly do I need to set up differently? Is it about where the model file is, or something about my phone's storage/filesystem?

<img width="1216" height="2688" alt="Image" src="https://github.com/user-attachments/assets/aa28800c-d949-4aef-95e9-9bfe0831c1e2" />

Helldez:
It depends on where the file is located, not on the phone: `/sdcard` and `Android/data` rely on a workaround. The engine performs a check at startup and falls back to an alternative mode. If you download the file via the app's catalog (or by using the "Select file" function), it is saved in the app's internal storage, where it works correctly.

M4RCORENE:
"Update: I found the issue myself — I was importing models via a file manager instead of adding them directly in the app. Once I did that, Direct I/O kicked in properly!
Results with Qwen3-235B-A22B (Q4_K_M):
- Before (buffered): 0.41 tok/s
- Now (O_DIRECT): 0.68 tok/s
That's a solid ~66% improvement, though not quite double — still, really happy with it!
[...]

PR #161, opened 2026-08-06
feat(server): add bmoe-server HTTP binary with OpenAI-compatible API

Bjynt:
I wanted to run this with Pi in Termux, but seems to be missing the server, so..
Added a new bmoe-server binary that loads a model once and serves it over HTTP, preserving the Session-based SSD streaming benefits between requests.

Exposes:
- GET /v1/models
- POST /v1/completions (with SSE streaming via stream=true)
- POST /v1/chat/completions (with SSE streaming via stream=true)

All bmoe-cli streaming flags are supported (--moe-stream, --cache-mb, --prefetch, etc.).

Tested on-device with MiniCPM-V 4.6 (505M) and Ornith 35B (22GB Qwen35MoE, 256 experts, 8 used/token with SSD streaming).

!! Code implemented by Laguna-S-2.1 !!


Helldez:
Sorry, I let this sit way too long. I want to get it in: people keep asking for a server (#147 again) and your PR is the right starting point.

What I need:

- Rebase on main. It conflicts, and #169 touched `session.cpp` yesterday.
- Why does the server change `config.h`, `session.h`, `runtime.cpp` and `session.cpp`? The CLI only uses the public `Session` API, and I'd like the server to do the same. If something is missing there, we add it as its own small change.

If you don't have time anymore, no problem: tell me and I'll finish it myself, your commits stay yours.

Issue #163, opened 2026-08-22
Fresh clone: doc-invoked scripts are not executable; CMake version trails the 0.20.0 release


gjjkbssg:
Two small build-integrity findings from setting up a fresh clone on a Mac (build and all 10 ctest gates pass there, for what it is worth):

**1. `scripts/build-host.sh` and `scripts/bench-run.sh` are committed as 100644.**

The README (Build from source), AGENTS.md, CONTRIBUTING.md and docs/seam.md all instruct running `scripts/build-host.sh` directly, and docs/benchmarks.md + docs/benchmark-method.md invoke `scripts/bench-run.sh` the same way. On any POSIX fresh clone this fails with `permission denied` (reproduced from a clean clone; everything works via `bash scripts/build-host.sh`). All `.sh` files in the tree carry the same mode — presumably invisible in a Windows working tree where `core.fileMode` is off, which would explain why it never surfaced.

Fix is mechanical: set the executable bit on the two doc-invoked scripts (PR incoming). The remaining `.sh` drivers are not invoked directly by the docs, so I left them alone — happy to flip those too if you prefer the whole set consistent.

**2. `CMakeLists.txt` still says `VERSION 0.19.0` while `[0.20.0]` is released (2026-08-17).**

Since `BMOE_VERSION` is injected from the CMake project version (core/CMakeLists.txt calls it the one source of truth), a build from the 0.20.0 tag self-reports as 0.19.0. I did not bundle the bump into the exec-bit PR to keep it focused — say the word and I'll send it separately, or fold it into your next release commit as you prefer.

Helldez:
Both confirmed, thanks.

The exec bit stays yours in #164, with all five flipped as you offered.

The version went in with #166, since it was tangled with the release bump. It was worse than it
looks: `project(VERSION)` is the only source of `BMOE_VERSION`, so every 0.20.0 build
self-reported as 0.19.0, in `--version` and in the `engine=` line of every CSV it wrote. Every
benchmark file committed since the tag names the wrong engine.

gjjkbssg:
Thanks — good to know the version half landed with #166, and agreed the CSV `engine=` corruption makes it release-blocking rather than cosmetic. The five-script extension is pushed to #164 and ready whenever you are.

Helldez:
Both halves are in now: exec bit with #164, version with #166. Thanks for both.

PR #164, opened 2026-08-22
Mark the two doc-invoked host scripts executable

gjjkbssg:
Fixes the exec-bit half of #163.

`scripts/build-host.sh` (invoked directly by README, AGENTS.md, CONTRIBUTING.md and docs/seam.md) and `scripts/bench-run.sh` (invoked directly by docs/benchmarks.md and docs/benchmark-method.md) were committed 100644, so the first documented step on a fresh POSIX clone fails with permission denied. Reproduced from a clean clone; `bash scripts/...` worked, which is why this hid in plain sight on a Windows working tree where the mode bit is invisible.

Only the mode changes — no content diff. The other `.sh` drivers are not invoked directly by the docs, so I left them for the issue discussion.

No CHANGELOG line: per AGENTS.md nothing accumulates outside a dated release section, and 0.20.0 is already cut — happy to fold a line into the next dated section if you want it recorded.

Verified on macOS 15 arm64 after the bit flip: `scripts/build-host.sh` configures and builds clean, and all 10 ctest gates pass (tiny-model byte-identity included).


gjjkbssg:
Extended to all five `scripts/*.sh` as you suggested in #166 — the commit is pushed. Leaving the set consistent so the next script cannot fail the same way behind a half-fixed tree.

Helldez:
That's the one, thanks. Five for five.

The version half of #163 landed on main with #166, so that issue closes once this does.

PR #167, opened 2026-08-25
feat(moe): warn when the expert cache budget sits below one token's cycle

### LINKED ISSUE #165: The cache-cliff guard the roadmap names: implemented on macOS, numbers inside
The roadmap closes the eviction-policy question negatively but keeps one guard on the list: *"the worst-case cycle is computable at init from model shape alone, so refusing or warning on a budget under it costs almost nothing"* (docs/roadmap.md, expert cache section; the trap is spelled out in docs/cache-sizing.md — 1500 MiB is a legal value and `--n-expert-used` moves the cycle past the floor silently).

I have it implemented and would like to agree on the shape before opening a PR:

- `ExpertStreamSource::worst_cycle_bytes(top_k)` — shape only: every bound layer's `entry_bytes(il) * min(top_k, n_expert)`, computed at init. No runtime demand measurement (that needs a decode to exist).
- One `fprintf(stderr)` from the session right after the streamer init, where both the final budget (auto or explicit) and the effective width (override or the model's own) are known. Warning, not rejection — the budget is legal and the run is byte-correct, just paying management and RAM for a cache that cannot hit.

Measured on the tiny qwen3moe gate model (macOS host build, 10/10 gates green with the change):

```
--cache-mb 1 --force-cache            → warns: below the 2 MiB worst-case token cycle (top-2)
--cache-mb 2 --force-cache (== cycle)  → silent
--cache-mb 4 (above)                   → silent
--cache-mb 0 (no LRU)                  → silent
--cache-mb 2 --force-cache --n-expert-used 4 → warns: below the 3 MiB cycle (top-4) — the k-override widens it, which is exactly the trap cache-sizing.md describes
```

(The k=4 cycle is 3 MiB, not 4, because entry_bytes is per-layer from the real tensor shapes — the guard prices from the model, not from a constant.)

Two open questions:

1. **Message wording** — it currently suggests `--cache-mb 0 or raise the budget`. cache-sizing.md argues `--cache-mb 0` is strictly better below the cliff; happy to reword or drop the advice half if you prefer the engine not editorialize.
2. **Where it belongs** — stderr next to the other `bmoe:` init lines is what I chose; `cli/main.cpp` owns env/UX, but it cannot see the auto-resolved budget. If you would rather have it in the summary line (`moe-cache: ...`) or as a metrics field, say so.

The diff is 3 files, +44 lines, no behavior change above the cliff. Branch is ready to push as soon as the shape has your nod.

gjjkbssg:
Closes #165. Related: #166 (which deliberately left this out so the guard lands as its own change).

Below one token's worst-case routed bytes, global LRU evicts precisely what it is about to read: the hit rate is 0% while the run still pays the cache's management time and its RAM. The only protection so far was `cache_min_mb`, a fixed floor that happens to sit above the cycle for the shipped models at their default top-k — and stops holding the moment `--n-expert-used` widens the routing without touching the budget.

**What this adds**

- `ExpertStreamSource::worst_cycle_bytes(top_k)` — the cycle priced at init from the model's shape alone: every bound layer's `entry_bytes` × `min(top_k, n_expert)`, at the top-k the run actually applies. Worst case, not the measured `token_demand_` (which needs a decode to exist).
- One stderr line at load when the resolved budget falls under it. It warns rather than refuses — the budget is legal and the output byte-identical — and states the fact without advice: the argument that `--cache-mb 0` is strictly better below the cliff stays in `docs/cache-sizing.md`, per your note on the original #166 draft.
- `cache_cycle_mb` in the metrics preamble, next to the budget it should be judged against (rounded up, from the resolved budget and the effective top-k) — so a committed CSV answers on its own whether its cache could ever have hit, without a second run of the same model and top-k. Documented in `docs/telemetry.md`.
- `docs/cache-sizing.md` and `docs/roadmap.md` updated in the same commit (the guard moves from "still worth doing" to shipped); CHANGELOG deliberately untouched since 0.21.0 is already cut — happy to fold the entry into your next dated section wherever you want it.

**Verification** (macOS arm64 host build)

The five cells from #165 reproduce identically:

| cell | result |
|---|---|
| `--cache-mb 1 --force-cache` | warns, cycle 2 MiB at top-k 2 |
| `--cache-mb 2 --force-cache` (== cycle) | silent |
| `--cache-mb 4 --force-cache` | silent |
| `--cache-mb 0` | silent |
| `--cache-mb 2 --force-cache --n-expert-used 4` | warns, cycle 3 MiB at top-k 4 — the k-override widens the cycle, exactly the trap cache-sizing.md describes |

The k=4 cycle is 3 MiB and not 4 because `entry_bytes` is per-layer from the real tensor shapes — the guard prices from the model, never from a constant. The preamble writes `... cache_ceil_mb=0 cache_cycle_mb=2 ...` next to the budget; 10/10 ctest host gates green.

Not yet validated on device — no engine hot path is touched (init-time arithmetic plus one `fprintf`), but a phone run before the next tag would close that. Also: formatted with clang-format 21 locally; if CI's pinned 18 disagrees anywhere I will follow up immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,24 @@ Semantic Versioning.
 
 ## [0.21.0] - 2026-08-25
 
+### Added
+- **The expert cache says when it cannot hit.** One token's worst-case routed bytes are priced at
+  load from the model's shape alone — every bound layer's expert-entry bytes times
+  `min(top_k, n_expert)`, at the top-k the run actually applies — and the engine prints one stderr
+  line when the resolved budget falls under it. Below that cycle global LRU evicts precisely what
+  it is about to read, so the hit rate is 0 % while the run still pays the management and the RAM.
+  It warns rather than refuses: the budget is legal and the output byte-identical, so the engine
+  states the fact and leaves the choice alone. The number is not the fixed `cache_min_mb` floor,
+  which is what made this invisible: a budget can clear the floor and still sit under *this* model's
+  cycle, and `--n-expert-used` widens the cycle without touching the budget. Validated on device on
+  Gemma 4 26B (cycle 902 MiB: 800 MiB measures a 0.0 % hit against 26.8 % at 950 MiB, decode
+  1.42 → 2.01 tok/s across that gap) and Qwen3.6-35B (cycle 582 MiB: 500 MiB measures 0.0 % against
+  36.0 % at 650 MiB). By @gjjkbssg (#165, #167).
+- **`cache_cycle_mb` in the metrics preamble.** The same number, recorded next to the budget it
+  should be judged against, so a committed CSV says on its own whether its cache could ever have
+  hit — no second run of the same model and top-k to compare with. See
+  [docs/telemetry.md](docs/telemetry.md) and [docs/cache-sizing.md](docs/cache-sizing.md).
+
 ### Changed
 - **Finer expert-cache rungs below 2000 MiB in the app.** The ladder went 500 → 1000 → 2000, and
   that x2 is where the choice is sharp: on an 8 GB phone 1000 MiB runs and 2000 MiB gets the app
--- a/core/include/bmoe/metrics.h
+++ b/core/include/bmoe/metrics.h
@@ -263,6 +263,12 @@ struct RunInfo {
     bool cache_auto = false;
     int cache_floor_mb = 0; // the RAM auto-sizing was told to leave free — the input behind cache_mb
     int cache_ceil_mb = 0;
+    // One token's WORST-CASE routed bytes, priced at load from the model's shape and the effective
+    // top-k. A cache_mb under this cannot hold a token cycle, so the run's hit rate is near zero by
+    // construction — and that is invisible in a committed CSV unless the cycle is recorded next to
+    // the budget, because the cliff is a property of the MODEL, not of the number that was typed.
+    // 0 when streaming is off.
+    int cache_cycle_mb = 0;
     bool force_cache = false;
     bool load_all = false; // whole-expert-set baseline: reads everything, so its bytes mean something else
     int io_threads = 0;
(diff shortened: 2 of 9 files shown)


Helldez:
Validated on a phone, on two real models rather than just the gate one.

Gemma 4 26B-A4B-Q4_K_M (17 GB, 128 experts, top-k 8), which exercises the fused gate_up plus shared-expert layout: cycle prices at 902 MiB, and 800 MiB measures 0.0% hit against 26.8% at 950 MiB. Decode goes 1.42 -> 2.01 tok/s across that 150 MiB, in a run where fmax stayed put.

Qwen3.6-35B-A3B-Q4_K_M: the arm64 build prices the cycle at 582 MiB, the same number the Windows host gives for that model, and 500 MiB measures 0.0% hit against 36.0% at 650 MiB. Throughput across that pair is not comparable, fmax moved between the cells, but hit rate does not care.

The five cells from #165 reproduce identically on arm64, and `--cache-mb auto` resolves and stays quiet above the cycle. Host side: gates 10/10 on Windows too.

Two things before merge:

- session.cpp: the cliff comment sits above the `if (route_trace)` block, so it reads as documenting the trace. Move it to the warning, or drop it since the warning carries its own.
- expert_stream_source.h: `cache_budget_bytes()` is never called, the check reads `im.info.cache_mb`. Please remove it.

CHANGELOG: 0.21.0 isn't tagged yet, only the bump and the section are on main. Put the entry in that section here.

gjjkbssg:
All three, in one amended commit:

- The stale cliff comment above the route-trace block is gone — the warning carries its own.
- `cache_budget_bytes()` removed; nothing calls it since the check moved to `im.info`.
- CHANGELOG entry added under the existing 0.21.0 section, with the on-device numbers from your validation folded in (Gemma 902 MiB cycle: 0.0% at 800 vs 26.8% at 950; Qwen3.6 582 MiB: 0.0% at 500 vs 36.0% at 650).

Thanks for the device runs — that the measured 0.0% lands exactly where the shape-priced cycle says it should, on both layouts and both arches, is the strongest evidence this guard could have.

PR #169, opened 2026-08-25
Fix telemetry attribution for critical-path expert stall

### LINKED ISSUE #98: Telemetry: compute is an unlabelled residual, so every run reads as compute-bound
The `compute_ms` residual makes almost every run read as "compute-bound", and the Android panel
amplifies it. v0.14.0 already recorded the consequence and left it open:

> `compute_ms` is a residual that has been absorbing zram decompression all along, so earlier
> "compute-bound" conclusions deserve re-examination.

This issue collects the three concrete follow-ups, plus the boundary of the claim.

## The asymmetry

Two of the three panel bars are measured (`stall_ms`/`io_ms`, `mgmt_ms`); `compute` is whatever
wall time is left, clamped at 0. So a large `flash wait` **proves** an I/O-bound token, while a
large `compute` proves nothing — it is also what you get when the cost is invisible to every
counter (zram swap-in, preemption, throttling). The panel can demonstrate one of the two verdicts
and not the other, which is why the compute-bound reading is the default outcome.

## Not everything here is an artefact

`docs/telemetry.md` (the `stall_ms` structural-floor note) already argues the other side: the
router only knows which experts a token needs *after* the gate runs, so the first miss can never be
hidden, and a warm run that reads as compute-bound is the **expected success case** — overlap did
its job. Measured floor there is ~5-15 ms/token at 76-81% hit.

So the ask is not "the panel is wrong". It is that the panel currently cannot distinguish
*successful overlap* from *unattributed cost*, and three specific things make that worse.

## 1. The app inverts the residual at end of run, against the documented contract

`docs/telemetry.md` states it explicitly:

> a consumer that recovers the flash-wait term as `wall_ms − compute_ms − mgmt_ms` gets
> `wall_ms − mgmt_ms` when the clamp fires, over-attributing to flash. Read the wall-additive flash
> term straight from `io_ms` (serial) / `stall_ms` (overlap) instead of inverting the residual.

`examples/android/.../Telemetry.kt` (`breakdown()`) does exactly that once the summary arrives:
live it derives compute from a measured flash term, but at end of run it takes `avgComputeMs` as
given and computes `flashWait = wall − compute − mgmt`. The two branches disagree, and the
end-of-run one — the number on screen longest — is the one the contract warns against. Worse, the
per-token clamps only ever round compute *up* and never compensate, so the aggregate is biased.

`stall_s_tok` is already carried in `BMOE_DONE`, so this is a wiring fix, not new instrumentation.

## 2. `stall_ms` is divided by `n_threads`, which is not the critical path

`core/src/engine/session.cpp` computes the overlap stall as the summed stall nanoseconds divided by
`cfg.n_threads`. But `ExpertStreamSource::on_expert_ready` only accumulates on threads that
actually block — a thread arriving after the slice landed returns early and contributes nothing. So
a wait borne by one or two threads is divided by the full compute thread count, and the remainder
silently becomes "compute". The docs concede the term is "a per-thread mean, not a critical path";
this asks for the critical path instead.

Suggested shape: an atomic count of currently-stalled threads; start a clock on the 0->1
transition, accumulate on the 1->0 transition. That measures the wall time the graph was actually
stalled, which is the wall-additive quantity the panel needs. Since threads synchronise on the node
anyway, one thread blocked means the graph is not progressing.

## 3. Nothing on screen is labelled "unattributed"

Proposal: four bars instead of three, none of them an unlabelled leftover.

| bar | source |
|---|---|
| compute | `cpu_ms / n_threads` — already collected, not currently used for the bars |
| flash wait | `stall_ms` (fixed per #2) |
| cache mgmt | `mgmt_ms` |
| **unattributed** | `wall − (the other three)` |

The fourth bar is the off-CPU time nobody claims: dense faults, zram, preemption. Today it is
painted as compute and read as "the model is thinking".

Related: the diagnostic line's "CPU % busy" stays misleading while the 2048-iteration
`std::this_thread::yield()` spin in `on_expert_ready` counts as busy CPU — it is waiting that looks
like work. Either shorten the spin or compute occupancy net of the stall term.

## Falsification

There is already a control case. The pinned-dense A/B measured +17.9% throughput with major faults
*equal* between the arms, and the whole delta landed in `compute_ms` (298 -> 241 ms) while `io_ms`,
`stall_ms` and hit rate moved less than 1%. Re-run that pair with the new bars: the
**unattributed** bar should drop by roughly that margin while **compute** stays flat. If it does
not, this diagnosis is wrong, and we find out from data already in the repo rather than from
another bench session.

Note that `--compute-trace` already "turns `compute_ms` from a residual into a measurement"
(v0.8.1), so it can arbitrate any disagreement — at a cost that keeps it a diagnostic, not
telemetry.

## Scope

Telemetry attribution only. No change to the streamer, the cache, or the seam; no effect on the
byte-identity gates. `docs/telemetry.md` ships with it.

Battery temperature in the metrics is tracked separately.

gjjkbssg:
Fixes #98

**Problem**

Overlap `stall` was the summed per-thread block time divided by the compute-thread count — a mean, not the critical path: whenever a minority of threads did the waiting, the stall was understated and the difference silently became "compute". The Android summary made it worse by reconstructing flash wait as `wall − compute − mgmt`, against the documented contract, so unmeasured runtime (zram swap-in, preemption, faults) was consistently mislabeled.

**Fix**

- Overlap stall is now the **union of stalled intervals**: the cumulative wall time during which at least one compute thread was blocked on a streamed expert. The interval opens on the 0→1 waiter transition (before the ready-spin — the need is already unmet) and closes on 1→0, with the transition and its timestamp taken inside one short critical section that holds no other lock and does no I/O. A snapshot taken mid-stall includes the open interval.
- The pure state machine (`StallUnionState`) is exercised deterministically with injected timestamps — overlap, nesting, separation, the 0→1/1→0 boundaries and the open-interval snapshot rule; the mutex wrapper owns the clock.
- `compute_ms` / `compute_s_tok` **retain their existing legacy residual semantics** for protocol and benchmark compatibility; only the Android attribution panel switches its Compute component to process CPU time normalized by compute-thread count.
- The Android panel draws four bars: **Compute** (`cpu ÷ computeThreads`, an attribution proxy, not a direct kernel measurement), **Flash wait** (measured `io` in serial / `stall` in overlap — in the end-of-run summary too, wired from `io_s_tok` / `stall_s_tok`, no residual inversion), **Cache mgmt**, and the non-negative **Unattributed** remainder. A missing measurement contributes 0 and its time stays unattributed; the bars are deliberately not rescaled to total 100%. The CPU-busy diagnostic keeps its own busy-thread denominator (compute + I/O lanes), which is a different quantity from the compute bar's.
- `docs/telemetry.md`, `MetricFields` and the CHANGELOG updated in the same commit; the spin loop, `ready_waiters_` lost-wakeup protocol and every streaming behavior are untouched.

**Validation**

- 11/11 host tests pass, including the overlap byte-identity gates (G4a–d, S2) — the new accounting sits on the expert-ready path and changes no bytes.
- `tests/stall_union_test.cpp`: deterministic interval-union cases (overlapping / nested / separated / boundary transitions / open-interval snapshot).
- clang-format 18.1.8 clean on all touched files.
- Android build not run locally (no SDK on this machine — the repo's Android path is Windows PowerShell + CI); the Kotlin changes are wiring-level and need a CI build.
- **Physical-device falsification was not run locally** because an Android device was unavailable. The natural re-test is the A/B from the issue: a run pair where the old panel booked a throughput delta as "compute" — under the new bars that delta should move to *unattributed* while compute stays flat.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [0.23.0] - 2026-08-27
+
+### Changed
+- **Telemetry attribution stops calling unmeasured runtime "compute".** Overlap `stall` is now the
+  **union of stalled intervals** — the cumulative wall time during which at least one compute thread
+  was blocked on a streamed expert — instead of the summed per-thread block time divided by the
+  thread count, a mean that understated the stall whenever a minority of threads did the waiting and
+  quietly dumped the difference into the `compute_ms` residual. The residual `compute_ms` /
+  `compute_s_tok` fields keep their existing meaning for protocol compatibility. The app panel now
+  draws four bars — compute (process CPU time over the compute threads, an attribution proxy),
+  flash wait (measured `io`/`stall`, in the end-of-run summary too: `io_s_tok`/`stall_s_tok` instead
+  of inverting the clamped residual), cache mgmt, and **unattributed** (the off-CPU wall time — zram,
+  preemption, faults — the residual used to absorb).
+- **Compatibility note:** under `--overlap`, `stall_ms` and `stall_s_tok` changed meaning — they are
+  the critical-path union now — so these columns are **not comparable with CSVs produced by older
+  releases**; compare them only within one engine version. Fixes #98.
+
 ## [0.21.0] - 2026-08-25
 
 ### Added
--- a/core/include/bmoe/expert_source.h
+++ b/core/include/bmoe/expert_source.h
@@ -72,7 +72,10 @@ public:
         long long cache_hits = 0;          // expert lookups served from the cache
         long long cache_lookups = 0;       // total expert lookups (hits + misses)
         uint64_t cache_resident_bytes = 0; // currently resident cached slice bytes
-        double stall_seconds = 0.0;        // overlap: summed across compute threads (0 when serial)
+        double stall_seconds = 0.0;        // overlap: cumulative wall time during which at least one compute
+                                           // thread was stalled on a streamed expert — the UNION of stalled
+                                           // intervals, not a per-thread sum (0 when serial). Includes an
+                                           // interval still open at the snapshot.
         uint64_t spec_read_bytes = 0;      // bytes read speculatively by prefetch (subset of read_bytes)
         long long spec_experts = 0;        // experts fully prefetched
         long long spec_useful = 0;         // prefetched experts that a later lookup actually hit
--- a/core/src/engine/session.cpp
+++ b/core/src/engine/session.cpp
@@ -98,7 +98,6 @@ llama_token argmax(const float * logits, int n_vocab) {
 struct GenTally {
     // Fixed for the run; kept here so record() needs only the token's own measurements.
     bool overlap = false;
-    int n_threads = 1;
 
     long long prev_bytes = 0;
     double prev_io_s = 0.0;
@@ -163,7 +162,11 @@ struct GenTally {
         m.io_ms = (st->read_seconds - prev_io_s) * 1000.0;
         m.mgmt_ms = (st->mgmt_seconds - prev_mgmt_s) * 1000.0;
         if (overlap) {
-            m.stall_ms = (st->stall_seconds - prev_stall_s) * 1000.0 / n_threads;
+            // stall is already wall-additive (the union of stalled intervals), so no thread-count
+            // normalization: dividing summed thread time by n_threads was a mean that understated
+            // the stall whenever a minority of threads did the waiting, and the difference quietly
+            // became "compute".
+            m.stall_ms = (st->stall_seconds - prev_stall_s) * 1000.0;
             m.compute_ms = m.wall_ms - m.stall_ms - m.mgmt_ms;
         } else {
             m.compute_ms = m.wall_ms - m.io_ms - m.mgmt_ms;
@@ -1143,7 +1146,6 @@ RunResult Session::generate(const GenerateRequest & req,
     // counters carry the prior prompts' totals; the deltas make each prompt self-relative.
     GenTally tally;
     tally.overlap = moe.overlap;
-    tally.n_threads = im.cfg.n_threads;
     if (moe.enabled) {
         const IExpertSource::Stats st0 = im.source.stats();
         tally.prev_bytes = (long long) st0.read_bytes;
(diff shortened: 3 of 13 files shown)


Helldez (changes requested):
Solid work, thanks. Right quantity, well tested, honest about what wasn't run. Two things before merge:

- CHANGELOG: the entry sits inside the released `[0.21.0]` section. Please move it to a new section for the next version.
- Add one line there: `stall_ms` / `stall_s_tok` under `--overlap` are not comparable with older CSVs, the column changed meaning.

Nit, optional: `exit_at` on an empty state drives `stalled_` to -1, and the next enter/exit pair accumulates `now - 0`. Can't happen today, but an early return on 0 closes it for free.

I'll run the A/B you describe on the reference phone before the next release. Sorry for the slow reply, the Qwen3.8-Flash-Next release ate the last two weeks.

gjjkbssg:
Thanks — addressed both merge items. The changelog entry now sits under a new `[0.23.0]` section (dated today, following the convention of the first PR into a section carrying the date — renumber freely if your next release lands as something else), with an explicit compatibility note that under `--overlap` `stall_ms` / `stall_s_tok` changed meaning and aren't comparable with older CSVs.

I also picked up the optional empty-state `exit_at` guard and added a regression case for it (a stray exit followed by a normal enter/exit must still produce the exact interval).

And thanks for offering to run the reference-phone A/B — the pinned model is bartowski's `Qwen_Qwen3.6-35B-A3B-Q4_K_M.gguf` from the app catalog, if useful for comparability.

Helldez:
Ran it on the reference phone (Qwen3.6-35B-A3B Q4_0, 4 threads, 4 I/O lanes, overlap). Steady-state ms/token:

| cache | compute (cpu/4) | stall | mgmt | wall | unattributed |
|---|---|---|---|---|---|
| 1750 MiB, 56% hit | 76 | 23 | 5 | 121 | +17 |
| 500 MiB, 0% hit, 546 MiB/tok | 175 | 185 | 59 | 345 | **-74** |

The union stall reads well: it crosses compute exactly where the run turns flash-bound, and tok/s is unchanged (8.2 both APKs at 1750). One thing to note: the `cpu / computeThreads` proxy includes the I/O lanes' CPU, and at 500+ MiB/tok that is large enough to push compute + stall past the wall; the panel clamps unattributed to 0 so it isn't visible, but the compute bar is overstated under heavy streaming. Worth saying in `docs/telemetry.md` that the proxy is an upper bound, or discounting the lanes if you see a clean way.

Helldez:
I don't see the push yet, the PR head is still 595078c. Once it's up I'll merge; the phone A/B above was on Q4_0, and the numbers hold.

Helldez (approved):
Verified on the reference phone, all points addressed. Merging.

PR #170, opened 2026-08-25
feat(moe): minor-fault telemetry, and a slot bank the device measured as negative

Helldez:
> **Draft, and not proposed for merge as it stands.** The device measured the slot bank as a
> negative on Android. What is worth keeping here is the telemetry and the demonstrated seam; the
> flag itself is parked. The history of this branch, including a wrong claim it carried for an
> hour, is left intact because the correction is the useful part.

## What this is

Two things: a metric that exposes a cost the streaming path was paying invisibly, and an opt-in
cache mode that removes it. The metric is the keeper. The mode works exactly as designed and still
loses on a phone, for a reason worth writing down.

## The cost, and why nothing could see it

The expert cache is direct-mapped by construction. `mul_mat_id` reads expert `e` at
`data + e*nb2`, so every expert has one address, and the only way to hold a byte budget is to
release the pages of whatever gets evicted. The next miss then reads into an address the kernel
must hand back and **zero** first, and the read overwrites every byte of that zeroing.

None of it reaches `mgmt_ms`. On POSIX `vm_commit` is a no-op, so the page materialises later, on
first touch, inside the reader, and bills `io`/`stall`/`compute`.

`minflt` and `minflt_mib` now travel with every token and in the run summary. Read against
`read_bytes`, the ratio is the finding, and it is the same on both platforms: **1.04 minor faults
per 4 KiB streamed** on device, 1.00 on host. Every streamed byte lands in a page zeroed for it
first.

## The lever

`--cache-slot-bank` gives each layer slots whose pages are claimed once and never released.
Eviction becomes a reassignment; a miss overwrites bytes already ours. What buys that is rewriting
the router's selected-expert ids to slot indices, legal because the kernel addresses `src0` as
`data + id*nb2` and never asks what the id means. The rewrite lands at the one node where no other
consumer is left to mislead: the terminal of the layer's weight chain, after `ggml_get_rows` has
taken the routing weights and before the first expert matmul.

No llama.cpp patch. This rides the existing eval-callback seam and the `tensor->data` rebinding the
streamer already does. It is a switch, not a size: the slot count is derived at load from the
budget and the per-expert bytes, and the bank comes out of `--cache-mb` rather than adding to it.

Constraints the engine enforces rather than documents: decode only (one `mul_mat_id` serves a whole
batch, so a prefill batch would need its entire expert set banked at once); slot count capped at
`n_expert`; and architectures whose experts carry a per-expert bias or scale read the same ids
after the matmul, so the bank disarms itself.

## What the device measured

Qwen3.6-35B-A3B Q4_0, 3000 MiB budget, in-app, six runs:

| | minflt/tok | **majflt/tok** | swap | hit | read | mgmt | tok/s |
|---|---:|---:|---:|---:|---:|---:|---:|
| baseline | 14 454 | **4-116** | ~100 MiB | 73.4 % | 54.4 MiB | 6.4 ms | 7.69-7.73 |
| slot bank | 615 | **1 676-5 484** | 155-393 MiB | 70.0 % | 63.7 MiB | 0.9 ms | 4.35-8.06 |

The mechanism works: 96 % of the minor faults gone, `mgmt_ms` down to noise, output
byte-identical. It loses anyway, and `majflt` says why.

The bank is permanently dirty anonymous memory, and with a working set of `top_k` slots out of the
27 it holds, most of it is untouched at any instant. That is precisely the profile Android's
`swappiness 160` compresses into zram. The pages we refused to hand back get taken anyway and come
back as major faults. Throughput tracks the swap directly: 8.06 tok/s at 155 MiB of swap, 4.35 at
393 MiB.

`rss_anon` was *lower* with the bank (2875-2949 vs 3073), so this is not about asking for too much
memory. It is about offering the kernel a large pool of lukewarm pages.

Which inverts the premise the branch was built on. **The decommit the bank removes is not waste.**
It is how the cache tells the kernel it genuinely does not need those pages, and that is what keeps
it out of reclaim. The baseline's 14 454 cheap minor faults per token buy it 4 major ones.

## A correction

An earlier revision of this branch claimed +8.2 % on the host. That number was taken with the bank
allocated *on top of* `--cache-mb` instead of out of it, so it was measured at 2.15x the baseline's
memory. Corrected sizing gives +1.8 % on the host at equal memory, and the device figures above.
The bank was never faster; it was bigger. Fixed in 07045c1, and the trap is documented in
`docs/cache-sizing.md` so the next reader does not repeat it.

## Two design faults still in it

- **Uniform slots per layer is a worse policy than a global LRU.** Every layer gets the same count
  whether it reuses experts or not. Hit rate 73.4 % to 70.0 %, which is 17 % more flash read.
- **The disarm check is broader than the danger.** It matches any MoE node named
  `_biased`/`_scaled`, catching the harmless router bias and routing-weight scale along with the
  genuinely unsafe per-expert ones. Combined with a graph counter that waits for layer 0's top-k,
  which never appears where there are leading dense blocks, the bank silently never arms on
  DeepSeek V4, Ling 3.0 or LFM2.

## Verification

- Byte-identity gates 10/10; generated text identical to baseline over 128 tokens
- Android cross-build clean (the only build that compiles the POSIX half of `platform_io.cpp`)
- clang-format 18 clean
- Device runs in-app on the branch build

## Docs

`CHANGELOG.md`, `docs/telemetry.md` (the new columns), `docs/cache-sizing.md` (the mode, its
limits, and the measured verdict), `examples/android/README.md`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,40 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [Unreleased]
+
+### Added
+- **The expert cache's page cost is now measurable, and optionally removable.** Two changes that
+  belong together.
+
+  `minflt` / `minflt_mib` join the per-token CSV and the run summary. Minor faults were the one
+  large cost in the streaming path that no metric could see. Evicting an entry releases its pages,
+  so the next miss reads into an address the kernel must hand back and zero before the read lands,
+  and none of that is billed to `mgmt_ms`: on POSIX `vm_commit` is a no-op, so the page arrives
+  later, inside the reader, and surfaces as I/O or compute. Read against `read_bytes`, the ratio is
+  the finding. On Qwen3.6-35B at a 2000 MiB budget it is 1.00, meaning every streamed byte lands in
+  a page zeroed for it first. Measured on the Windows host too, where the cache decommits and
+  recommits explicitly.
+
+  `--cache-slot-bank N` is the lever that answers it. Each layer gets N slots whose pages are
+  claimed once at load and never released, so eviction is a reassignment and a miss overwrites
+  bytes already ours. It works by rewriting the router's selected-expert ids to slot indices, which
+  is legal because `mul_mat_id` addresses `src0` as `data + id*nb2` and never asks what the id
+  means. The rewrite happens at the single node where no other consumer is left to mislead: the
+  terminal of the layer's weight chain, after the routing weights have been gathered and before the
+  first expert matmul. Decode only, since one `mul_mat_id` serves a whole batch. N is capped at the
+  model's expert count. Architectures whose experts carry a per-expert bias or scale read the same
+  ids after the matmul, so the engine detects them from the graph's node names and disarms itself.
+  Off by default, and off in the app.
+
+  Host result on Qwen3.6-35B, three A/B pairs: 2.805 to 3.035 tok/s (+8.2 %), minor faults 61 323
+  to 0 per token, `mgmt_ms` 31.1 to 0.9 ms, generated text byte-identical. The attribution is not
+  the one the mechanism suggests, and is documented as such: ms per MiB read did not move, so the
+  zeroing was already hidden behind that machine's SSD wait. What paid was the eviction syscalls
+  disappearing, plus a hit-rate rise (54.9 to 57.5 %) that falls out of the bank's per-layer LRU
+  giving every layer a floor. A phone inverts both halves of that balance, which is why nothing is
+  turned on until it has been measured there. See `docs/cache-sizing.md` and `docs/telemetry.md`.
+
 ## [0.21.0] - 2026-08-25
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,24 +19,31 @@ Semantic Versioning.
   a page zeroed for it first. Measured on the Windows host too, where the cache decommits and
   recommits explicitly.
 
-  `--cache-slot-bank N` is the lever that answers it. Each layer gets N slots whose pages are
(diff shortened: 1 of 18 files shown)


Helldez:
**Closing this as rejected.** The device settled it, and the verdict is worth stating plainly rather than leaving a draft open.

The slot bank does exactly what it was designed to do — 96% of the minor faults gone, `mgmt_ms` down to noise, output byte-identical — and it is still a regression on Android. The reason is the one recorded above: the bank is a large pool of permanently-dirty, mostly-lukewarm anonymous pages, and `swappiness 160` compresses precisely that profile into zram. Throughput tracks the resulting swap directly (8.06 tok/s at 155 MiB of swap, 4.35 at 393 MiB), and `rss_anon` was *lower* with the bank, so this is not a matter of sizing it down.

That inverts the premise the branch was built on. The decommit the bank removes is not waste: it is how the cache tells the kernel it genuinely does not need those pages, and that is what keeps the cache out of reclaim. Trading ~14.5k cheap minor faults per token for a handful of major ones is a bad trade on this platform, and no amount of tuning the slot policy changes the sign — the two design faults listed above (uniform per-layer slots losing 3.4pp of hit rate, and the over-broad disarm check) would each cost more work than the mechanism could ever return.

Not reopened without a platform where anonymous memory is not reclaim bait.

The `minflt` / `minflt_mib` columns go with it. The finding they carry — 1.04 minor faults per 4 KiB streamed on device, 1.00 on host, i.e. every streamed byte lands in a page zeroed for it first — is real and independent of the mode, but it is telemetry for a cost we have now decided not to pay down, so it is not worth carrying in `main` on its own. It stays here as evidence.

One trap is worth remembering even outside this branch: a cache experiment must be sized *out of* the existing budget, never on top of it. Measured the other way, this mode read as +8.2% on the host when the honest figure at equal memory is +1.8%.

The branch is deleted; both commits stay reachable from this PR if the mechanism is ever worth revisiting.

PR #173, opened 2026-08-26
docs(roadmap): state the prefill objective

Helldez:
Adds one roadmap theme: making prefill fast.

Every lever on the roadmap so far was chosen, built and judged against decode. Prefill runs on the
defaults and has never been sized. On a >RAM MoE that is what the user actually waits for: a 1334
token prompt on Qwen3.6-35B-A3B measures 11.4 tok/s on the reference device, about two minutes
before the first token appears, ahead of a generation that then runs at 5 tok/s.

The theme also records why nothing is proposed yet. `Metrics` has `prefill_seconds` and `n_prompt`,
but the stall and byte counters that separate compute from flash are per decode token, so prefill's
cost is unattributed, and the two figures the repo does hold disagree about which of the two binds.

Objective only. No lever, no plan, no code.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -10,6 +10,24 @@ So the streaming path has already recovered most of what streaming can recover, 
 below are ordered by that fact: read bandwidth still matters for the models that do not fit a
 useful cache, but throughput on the ones that do now depends on compute.
 
+## Prefill, and the half of the run nobody measured
+
+Every lever on this page was chosen, built and judged against decode. Prefill got the defaults:
+`n_batch = n_ctx`, `--ubatch 512` in the app, dropping armed for generation only. Nothing here was
+ever sized against it.
+
+That is a gap in the product, not only in the docs. What a user waits for is
+`load_seconds + prefill_seconds`, and on a >RAM MoE the second term runs in minutes: a 1334 token
+prompt on Qwen3.6-35B-A3B measured 11.4 tok/s on the reference device, roughly two minutes before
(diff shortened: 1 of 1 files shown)


gjjkbssg:
Hey, I've been looking into this while #169 is pending. I ran a CPU-only host reconnaissance on Qwen3.6-35B-A3B and found that the existing cumulative streamer counters are already enough to attribute prefill at the session layer without changing the streamer.

As one structural data point, unique `(layer, expert)` coverage grew from 40.2% at ~130 prompt tokens to 77.0% at ~1658 tokens, so the prefill working set keeps expanding but starts to saturate rather than sweeping the whole bank immediately. The host timing data also suggests the storage/compute balance changes with prompt length, though I'd treat that as a hypothesis until it's checked on the reference device.

If this direction makes sense, I'd be happy to take a focused follow-up that adds prefill-phase attribution (wall / CPU / read bytes / I/O / stall / mgmt) using session-level deltas once #169 lands.

Helldez:
Yes, that's the right direction. Go ahead once #169 is in:

- session-level deltas, streamer untouched
- host measurements in the PR itself

The coverage curve is a useful data point on its own. The two prefill figures in the repo disagree about what binds; a per-phase split is what would settle it.

gjjkbssg:
Got it — session-level deltas with the streamer untouched, and the host measurements in the PR itself. Starting as soon as #169 lands.

Helldez:
#169 is in. Whenever you're ready.

Helldez:
Reopened: closed by accident by closing #175. Nothing here was decided or rejected.

PR #175, opened 2026-08-27
feat(telemetry): add prefill-phase attribution

gjjkbssg:
Follow-up to #173.

The prompt phase of a >RAM run is what the user actually waits for before the first token, but today it carries only `prefill_s`, a bare wall-clock measurement. The streamer counters that separate CPU activity, reads, I/O wait, stall, and management are cumulative, while the decode baseline is deliberately taken after prefill. That keeps decode telemetry clean, but leaves prefill itself unattributed.

This adds a per-phase split so the two existing prefill observations in the roadmap can be tested against the same quantities.

**Change**

Session layer only; the streamer is untouched.

* Snapshot the existing cumulative streamer counters and process CPU time immediately before and after this turn's prefill chunks.
* Append five fields to `BMOE_DONE` alongside the existing `prefill_s`:

  * `prefill_cpu_s` — aggregate process CPU seconds during prefill.
  * `prefill_read_mib` — streamer bytes read during prefill.
  * `prefill_io_s` — summed I/O-lane busy time.
  * `prefill_stall_s` — critical-path streamed-expert stall, using the overlap interval-union semantics from #169.
  * `prefill_mgmt_s` — streamer/cache management time.
* The existing post-prefill snapshot remains the decode baseline, so decode telemetry is unchanged.
* The keys are additive; existing parsers can ignore fields they do not know.

`prefill_cpu_s` is deliberately reported as aggregate process CPU time, not as "compute". Under overlap it includes CPU consumed by the I/O lanes and other process threads, matching the limitation documented in #169.

**Host measurements**

Qwen3.6-35B-A3B, bartowski Q4_K_M; CPU-only build (`GGML_METAL=OFF`); macOS arm64; external SSD; 4 compute threads / 4 I/O lanes; overlap; 3000 MiB cache.

This is architecture reconnaissance, not a reference-device benchmark. macOS has no O_DIRECT here, so `read MiB` means streamer bytes read, not guaranteed physical-device bytes.

For comparison only, the CPU and I/O columns below normalize the raw aggregate counters by their four threads/lanes:

| prompt tokens | wall s | stall s | cpu/4 s | io/4 s | streamer read MiB | unique `(layer, expert)` coverage |
| ------------: | -----: | ------: | ------: | -----: | ----------------: | --------------------------------: |
|           130 |   36.0 |    25.3 |     3.4 |   26.0 |            10,200 |                             40.2% |
|           261 |   42.7 |    31.7 |     5.5 |   33.5 |            12,907 |                                 — |
|           448 |   54.1 |    35.8 |     8.2 |   39.5 |            14,419 |                             61.2% |
|           988 |   69.6 |    33.0 |    16.0 |   40.4 |            15,933 |                                 — |
|          1249 |   78.3 |    32.7 |    20.2 |   40.8 |            16,196 |                             73.9% |
|          1658 |   70.1 |    26.3 |    25.6 |   41.7 |            16,505 |                             77.0% |

Coverage was computed offline from `--route-trace`; it is not part of the production telemetry added here. Run-order / host variance is visible in the wall column, so the shape rather than any individual timing cell is the useful observation.

On this host, the split suggests that short prompts spend a large share of wall time waiting for streamed experts. As unique `(layer, expert)` coverage expands and begins to saturate, stall falls after its peak while the normalized process-CPU contribution continues to grow. The two are close by roughly 1.6k prompt tokens in this host run.

That is a host-side hypothesis, not a device conclusion. The purpose of these fields is to make the same split directly measurable on the reference phone and determine where — or whether — that crossover occurs there.

**Validation**

* 11/11 host gates pass.
* clang-format 18.1.8 clean.
* A real Qwen3.6-35B-A3B `--session` run emits all five new `BMOE_DONE` fields.
* The emitted prefill read / I/O / stall / management / process-CPU deltas match the independent session-only reconnaissance probe for the same run.
* Streamer code and decode attribution are unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project are documented here. The format follows
 [Keep a Changelog](https://keepachangelog.com/), and the project aims to follow
 Semantic Versioning.
 
+## [0.23.0] - 2026-08-28
+
+### Added
+- **Prefill-phase attribution in `BMOE_DONE` and the CSV `# summary`.** `prefill_cpu_s` /
+  `prefill_read_mib` / `prefill_io_s` / `prefill_stall_s` / `prefill_mgmt_s`: the prompt phase's
+  own split of the same cumulative counters the decode fields come from, as session-level deltas
+  across the prefill chunks — the streamer is untouched. On a >RAM model the prompt is what the
+  user actually waits for before the first token, and until now the phase carried a single bare
+  wall number (`prefill_s`) with no compute/flash split; these keys are what settles which of
+  the two binds at a given prompt length (#173). The CSV `# summary` trailer carries the same
+  five raw fields under the same names and units — appended keys, so existing readers ignore
+  them. Host measurements in the PR.
+
 ## [0.22.0] - 2026-08-28
 
 ### Added
--- a/cli/main.cpp
+++ b/cli/main.cpp
@@ -330,6 +330,8 @@ static int run_session_loop(const RunConfig & cfg,
                     "\"prefill_tps\":%.2f,\"load_s\":%.3f,\"cache_hit_pct\":%.1f,\"n_prompt\":%d,\"n_past\":%d,"
                     "\"compute_s_tok\":%.4f,\"io_s_tok\":%.4f,\"cache_resident_mib\":%.0f,\"cache_budget_mib\":%.0f,"
                     "\"read_mib\":%.1f,\"stall_s_tok\":%.4f,\"mgmt_s_tok\":%.4f,\"majflt_tok\":%.2f,\"cpu_s_tok\":%.4f,"
+                    "\"prefill_cpu_s\":%.3f,\"prefill_read_mib\":%.1f,\"prefill_io_s\":%.3f,"
+                    "\"prefill_stall_s\":%.3f,\"prefill_mgmt_s\":%.3f,"
                     "\"token_demand_mib\":%.1f,\"mtp_drafted\":%lld,\"mtp_accepted\":%lld,\"mtp_decodes\":%lld,"
                     "\"mtp_draft_s_tok\":%.4f,\"drafted_steps\":%lld,\"loop_overhead_s_tok\":%.4f,"
                     "\"reasoning\":\"%s\",\"text\":\"%s\"}\n",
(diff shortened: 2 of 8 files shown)


Helldez (changes requested):
Clean, and exactly what we agreed. Two things before merge:

- Rebase on main. 0.22.0 shipped this morning and the `[0.23.0]` section got folded into it, so your entry would land inside a released section. Put it in a new section for the next version.
- Add the five fields to the CSV `# summary` line too (`core/src/metrics/csv_metrics_sink.cpp:72`). That line is what the app writes on the phone and what I read back for the device check; without it I can't see your split there.

Then I'll run a short and a long prompt on the phone and merge.

Helldez:
One more, optional: the five prefill locals in `generate()` could be a small struct with a begin/end pair, like `GenTally` does for decode. That function is already long, and the delta logic would live in one place. Fine to leave it if you'd rather keep the diff small.

gjjkbssg:
Rebased onto the current main and moved the changelog entry into the new next-version section.

The CSV `# summary` now carries the same five raw prefill fields as `BMOE_DONE`, and I verified them end-to-end in the same session output. I also picked up the optional tally cleanup, so the prefill delta logic now lives in a small `PrefillTally` rather than directly in `generate()`. All local gates and CI checks are green.

Helldez (approved):
Both points addressed, and the CSV regression test was a good call. Device check done.

Ran it on the reference phone against Qwen3.6-35B-A3B (Q3_K_XL), with the engine executed under the app's uid from the installed APK's lib dir, so internal storage and real O_DIRECT: 4 compute threads, 4 I/O lanes, overlap, pinned dense, drop-cold 0.75, 1500 MiB cache, ctx 2048. Two turns in one session, a 73-token prompt then a 1214-token one. That is also the sharpest test of the delta logic, since a cumulative leak would put turn 1's totals inside turn 2.

| | turn 1 | turn 2 |
|---|---:|---:|
| prompt tokens | 73 | 1214 |
| `prefill_s` | 9.53 | 92.85 |
| `prefill_cpu_s` | 31.34 | 323.60 |
| `prefill_read_mib` | 6315.8 | 29132.9 |
| `prefill_io_s` | 9.94 | 45.28 |
| `prefill_stall_s` | 0.229 | 0.000 |
| `prefill_mgmt_s` | 0.67 | 4.27 |

All five keys land in both `BMOE_DONE` and the CSV `# summary`, one per turn. Across two independent runs `prefill_read_mib` came out identical to the decimal while every timing moved, which is what deterministic byte counters and real timings should each do.

The device answer differs from your host one, and that is the useful part. Divided by the four threads, turn 1 is 8.2 s of CPU against 10.4 s of wall, so 79% compute with stall at 3.4%, and the long prompt drives stall to exactly zero: the reads sit entirely behind compute. The crossover you place around 1.6k prompt tokens on host has already happened before 73 tokens here. Two caveats on my run, both of which argue the same direction rather than against it: the cache was 1500 MiB rather than the app's 3000 (at 3000 the process was reclaimed mid-turn, since a binary under `run-as` has none of the foreground app's protection), and the CPU was frequency-capped, which biases toward the compute-bound reading.

So the fields do exactly what they were added for: they turn #173's question into something you read off a run instead of infer. Merging now, no release tag yet.

Issue #179, opened 2026-08-28
Apple platform support: F_NOCACHE on macOS, and an iOS target


Helldez:
Two gaps in Apple support, one small and one large. Both are open to contributors, and both need a measurement, not just code.

## 1. macOS never bypasses the page cache

`core/src/io/platform_io.cpp` defines `O_DIRECT` to `0` when the platform lacks it, which on Apple means `open_read(..., direct=true)` silently returns an ordinary buffered descriptor. The equivalent there is `fcntl(fd, F_NOCACHE, 1)` after the open, and it is never called. So on a Mac the expert stream reads through the page cache, which is precisely what the design exists to avoid.

Second half of the same problem: the `o_direct` metrics field is filled from the *requested* configuration, so a macOS run reports `o_direct=1` while running buffered. The field should record what the open actually achieved, on every platform. That half is worth fixing on its own, even if nobody touches the macOS path.

**What a PR needs**

- `F_NOCACHE` applied on Apple where `O_DIRECT` would be used elsewhere, without disturbing the Linux, Android or Windows paths.
- `o_direct` reported from the open's real outcome rather than from config.
- An A/B on a real Mac: same model, same protocol, `F_NOCACHE` off against on, with the columns [docs/community-benchmarks.md](docs/community-benchmarks.md) asks for.

`F_NOCACHE` is not `O_DIRECT`: it does not require alignment and promises no DMA path, so it may well come out neutral or negative. That is a publishable result and exactly why the code alone is not enough to merge. If it does come out negative, the honest-reporting fix still lands and the caveat in [docs/limitations.md](docs/limitations.md) gets the measured number.

Good first issue if you have a Mac. Two files, plus a run.

## 2. There is no iOS target

The core is portable C++ and the streaming path has no Android dependency, but iOS does not run command-line binaries, so anything here means an app: an Xcode project or a Swift package wrapping the engine, plus somewhere for the model file to live.

Before writing much code, a proposal should answer:

- Does `F_NOCACHE` behave on iOS as it does on macOS?
- What does the per-app jetsam memory limit do to a bounded expert cache? Streaming keeps RSS low, which is in our favour, but the limit is hard and per-app.
- How does an 18 GB gguf get onto the device at all, and where does it live?

This one is a question, not a task. Comment with the shape you have in mind before opening a PR.

## Ground rules for both

[...]

PR #182, opened 2026-08-29
feat(io): use F_NOCACHE on macOS and report effective direct mode

gjjkbssg:
Addresses the macOS half of #179.

On Apple, `O_DIRECT` is not available. Until now a BigMoe direct-I/O request therefore opened an ordinary buffered descriptor while telemetry still reported `o_direct=1` from the requested configuration.

This patch separates two concepts that coincide on the existing direct-I/O platforms but do not on Apple:

* **effective uncached/direct mode** — whether the platform open actually achieved the requested uncached behavior;
* **alignment-constrained reads** — whether the platform requires the O_DIRECT / no-buffering aligned-read path.

On Apple, a direct request now opens the file normally and applies `fcntl(fd, F_NOCACHE, 1)`. A successful `F_NOCACHE` makes the descriptor effectively uncached for telemetry purposes, but it does **not** enable the O_DIRECT alignment/bounce path: reads remain ordinary `pread` operations with no new alignment contract or DMA claim.

Independently, `o_direct` now comes from the actual open outcome rather than the requested configuration. The result originates at the platform open, is preserved through `FileReader`, and is reported conservatively across expert shards: a run reports `o_direct=1` only when every relevant streaming reader achieved its platform uncached/direct mode. Existing refusal/verification fallback paths report `0` after falling back to buffered I/O.

Linux, Android and Windows retain their existing I/O behavior; this changes only the truth source for the metric.

## macOS A/B

CPU-only build (`GGML_METAL=OFF`), Qwen3.6-35B-A3B Q4_K_M, 16 GB Apple-silicon Mac, model on an external volume.

The current community benchmark protocol was used with one declared override: `CACHE_MB=3000` was held fixed because `auto` sizes from free memory at model load and would otherwise move between runs. All other settings, model, prompt, binary and path were identical.

Run order was interleaved:

`buffered → F_NOCACHE → F_NOCACHE → buffered → buffered → F_NOCACHE`

| mode        | decode tok/s | stall s/tok | legacy compute residual s/tok | I/O s/tok | cache hit | streamer MiB/tok |
| ----------- | -----------: | ----------: | ----------------------------: | --------: | --------: | ---------------: |
| buffered    |        0.605 |       0.433 |                         1.223 |     1.733 |     70.1% |              159 |
| `F_NOCACHE` |        0.941 |       0.462 |                         0.600 |     1.612 |     70.1% |              159 |

The three buffered cells were 0.582 / 0.586 / 0.646 tok/s; the three `F_NOCACHE` cells were 0.970 / 0.907 / 0.947 tok/s, a +55.6% mean decode improvement on this configuration with non-overlapping cells.

The routed byte stream was identical across all six runs (`read_MiB=40822.8`), as were the cache-hit rate and effective streamed MiB/token. Stall also remained similar. The observed throughput difference lands primarily in the legacy compute residual. That is consistent with reduced page-cache / memory-pressure overhead in this >RAM configuration, but the residual is intentionally not treated as a direct measurement of that mechanism.

Telemetry also verifies the intended mode in the real runs:

* explicit buffered arm: `o_direct=0`
* `F_NOCACHE` arm: `o_direct=1`

This is one machine, one filesystem/volume and one model; it is a measured result for this configuration, not a general macOS performance claim. `F_NOCACHE` is not Linux `O_DIRECT`, and other storage devices may be neutral or slower.

## Validation

* CPU-only macOS build passes.
* 12/12 host gates pass.
* clang-format 18.1.8 clean.
* Buffered and `F_NOCACHE` real-file smoke runs report the expected effective `o_direct` values in both the streaming banner and CSV.
* Final branch is rebased onto current `main`, including #175.
* Linux/Windows were not cross-compiled locally; CI will validate those platform paths.

Documentation updates the Apple limitation, telemetry meaning of `o_direct`, and the community benchmark note.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -83,6 +83,17 @@ Semantic Versioning.
 - `--cache-mb auto` no longer reserves the size of a row-streamed table for a conversion that will
   not happen. It deducts the window instead, so the cache gets the RAM the policy freed rather than
   the engine planning for both.
+- **macOS reads uncached (`F_NOCACHE`), and `o_direct` reports the open's real outcome everywhere.**
+  A direct request on Apple applies `fcntl(F_NOCACHE, 1)` to every reader descriptor — the kernel
+  stops caching that file's pages — instead of silently returning a buffered fd (Apple has no
+  `O_DIRECT`). It is a caching hint, not an I/O mode: no alignment contract, no DMA promise, so a
+  direct reader on Apple keeps ordinary `pread` semantics and skips the O_DIRECT bounce path rather
+  than inheriting Linux-only alignment requirements. Independently, the `o_direct` telemetry field
+  (CSV preamble, decode-trace header, streaming banner) now comes from what the shard opens
+  achieved — the AND across shards, after every platform refusal and open-time downgrade — instead
+  of from the requested configuration, so a run served buffered says `0` on every platform. Host
+  measurements in the PR (#179).
+
 ## [0.23.0] - 2026-08-29
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -401,8 +401,9 @@ The model must live on a real filesystem (on Android `/data/local/tmp/...`, not 
 Platform status: Linux is exercised by CI (build + gates) and Windows is where the
 [desktop numbers](#desktop) were measured. On Windows, build with CMake directly (Visual Studio
 Build Tools); the script above is bash, and MSVC puts the binary in `build\cli\Release\bmoe-cli.exe`.
-macOS builds from the same sources (the platform branches exist) but is not validated, and it has
-no O_DIRECT, so direct reads fall back to buffered I/O there.
+macOS builds from the same sources (the platform branches exist) but is not exercised by CI. It has
+no O_DIRECT; a direct request is served with `F_NOCACHE` instead (uncached, but not alignment-constrained),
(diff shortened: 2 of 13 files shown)


Helldez (commented):
Thanks I'm reviewing it
Great job

Helldez (approved):
Reviewed the whole diff and the callers. The split between "uncached descriptor" and "alignment-constrained reads" is clean, the platform difference stays inside platform_io where it belongs, and every fallback path (refused open, failed verify, F_NOCACHE refusal) ends up with a consistent reader. Nice work, and thanks for the interleaved A/B.

Three mechanical things I'll take care of myself before merging, so you don't need to touch the branch:

- main moved this morning (#183, #180, #171) and your CHANGELOG entry landed under 0.23.0, which is already tagged. I'll rebase onto current main and move it under `[0.24.0] - unreleased`. That's a force-push to your fork branch, heads up if you had local work on it.
- README still says macOS "has no O_DIRECT, so direct reads fall back to buffered I/O there". No longer true after this, I'll fix that line in the same rebase.
- I'll drop the `keepachangelog.com/` link touch, it's unrelated.

Not blocking, but a good follow-up PR if you feel like it: ci.yml has no macOS job, so the `__APPLE__` branch is only compiled by release-host at release time. A build-only `macos-14` job would catch a compile break at the right moment.

PR #185, opened 2026-08-29
feat(io): release the model file's mapping after load (--release-mmap)

Helldez:
## The symptom

On the desktop host, adding I/O lanes has never done anything. `--io-threads 4` and `--io-threads 8`
measure the same, the serial read path once measured *exactly* the one-lane rate with four lanes
open, and the engine has been reading at roughly a third of what the drive can serve at the request
shape it actually issues. Every previous investigation blamed the read path (duty cycle, the bounce
buffer, request scatter) and every fix for those was worth a few per cent at best.

None of it was the read path.

## The cause

llama.cpp maps every gguf it loads and keeps the mapping for the model's lifetime. That is
load-bearing for this engine: the streamer rebinds expert tensors onto the file's native byte
layout, which is why `use_mmap=true` is an invariant here. But on Windows, **while a section of a
file is alive, concurrent unbuffered reads on that file are serialised**, and a descriptor opened
*while* that section existed keeps serialising against it even after the section is gone.

So the four reader lanes were never reading in parallel. They were queueing.

Measured with `bmoe-iobench` at the engine's own request shape (576 KiB, 4 lanes, O_DIRECT, random
offsets) on an NVMe host, one variable per row:

| variable | 1 lane | 4 lanes |
|---|---:|---:|
| nothing (the drive itself) | 883-967 MiB/s | **2295-2665 MiB/s** |
| offsets confined to one region (`--range-mb`) | 917 | 2651 |
| lanes reading against a busy CPU (`--compute-load 8`) | n/a | 2463 |
| destination pages committed afresh per read (`--fresh`) | 745 | 1952 |
| **a read-only mapping of the file held open (`--mmap`)** | 909 | **895-930** |
| mapping dropped and lanes reopened (`--mmap --reopen-lanes`) | n/a | **2402** |

Locality, CPU contention and page-commit cost move the number by a few per cent. The mapping moves
it by **2.6x**, and under it four lanes deliver exactly one lane's throughput. Nothing is ever read
*through* the mapping in these cells: its existence alone is the variable.

Sub-cells that placed the blame precisely, all at 4 lanes: the mapping opened before the lanes
versus after them (927 MiB/s either way, so it is not open order); reading through a hard link while
the original is mapped (927, so it is the file, not the handle); unmapping the view but keeping the
section open (926) versus closing the section too (2050-2207, recovered).

## The fix

After load, nothing reads through that mapping any more: the dense weights have been copied into
the engine's own buffers and the experts are streamed from the engine's own descriptors. So with
`--release-mmap` the engine hands the mapping back. It unmaps the file, closes its section, and
**reopens the reader lanes**, which is the second half and is not optional, since dropping the
mapping alone leaves the lanes serialised.

Whether that is safe is decided by looking rather than by reasoning about which tensors ought to
have been rebound. The engine asks the OS whether any weight the capture pass observed still points
inside a mapping of the model files, and declines if any does. One question covers every residency
policy at once: a dense set deliberately left mmap'd, a table held back as oversized, or a tensor no
name-based accounting could have found. Running with `--dense-weights mmap` reports the count and
stands down.

llama.cpp is not patched and still believes it owns its mapping, so what its teardown will unmap and
close is left in place as an inert placeholder: an address reservation where the view was, and an
inert handle in the section's slot, both released only after the model is freed. On Windows this
surfaces as a `warning: UnmapViewOfFile failed` line at teardown, printed by llama.cpp. It is the
designed outcome, and `docs/moe-streaming.md` now says so, so nobody reports it as a bug.

## Results

**Windows host**, Qwen3.6-35B-A3B Q4_K_M, `--cache-mb 3000 --io-threads 4 --overlap
--dense-weights anon -t 8 --ubatch 256 -c 1024`, 96 tokens, cells interleaved:

| | baseline | `--release-mmap` |
|---|---:|---:|
| decode | 3.16 tok/s | **4.63 tok/s (+46%)** |
| flash stall | 0.182 s/token | **0.074 s/token** |
| per-read latency p50 | 2.49 ms | 1.20 ms |
| flash read per token | 192.5 MiB | 192.5 MiB |
| cache hit / evictions / re-reads | 60.9% / 11635 / 7848 | identical |
| generated text | baseline | byte-identical |

Not one byte fewer is read and no quality knob is touched. The same reads are simply served in
parallel.

**Android**, same model and device, `--cache-mb 2000 -t 4`, 48 tokens, two cells per variant in both
orders with a thermal gate: flash stall is **unchanged** (0.087-0.093 s/token either way), exactly as
the microbenchmark predicts, because f2fs does not serialise. The flag is still 5-9% faster there,
for a different reason: CPU time per token falls about 9%, because keeping a 20 GB mapping
registered is not free for a kernel already under memory pressure. Prefill, model load and TTFT are
unaffected. That measurement is two short cells per variant against a device whose prefill alone
spread 20% across identical cells, so it is recorded as a direction, not a number, and a 256-token
protocol is owed before it becomes a published figure.

The flag is **off by default** on both platforms pending that.

## What is in here

- `core/src/io/mapping_release.{h,cpp}`, the release itself, Windows and POSIX, with the placeholder
  discipline that makes llama.cpp's own teardown harmless.
- `FileReader::reopen()`, propagated through `ExpertStreamSource` (refused under the queue lock if
  any read is in flight) and `RowStream`.
- The safety decision in `Session::open`, and `--release-mmap` / `MoeStreamConfig::release_mmap`.
- The dense policy now records every distinct leaf tensor rather than one per name, and rebinds
  every tensor over one file range onto the same buffer. See the tied-head fix below.
- `tests/mapping_release_test.cpp`, which builds the exact post-load state llama.cpp leaves on a
  scratch file, releases it, and asserts both what must happen and what must be left behind,
  including that the model's later unmap and close are harmless. Runs on both platforms.
- `bmoe-iobench` grows `--mmap` and `--reopen-lanes`, which reproduce both halves of the mechanism
  in one cell each, plus `--range-mb` and `--fresh`, the two controls that ruled out locality and
  page-commit cost.
- `docs/bench-data/2026-08-29-mmap-serialisation/findings.md`, every cell above, the sub-cells that
  isolated the mechanism, and two hypotheses this work disproved along the way (the stall-accounting
  mutex costs 0.4%, inside noise; and a second engine on the same drive invalidated an earlier round
  of numbers).
- The demo app exposes it as **"Release the model mapping"**, enabled only under a dense policy that
  rebinds every weight into the app's own memory (Anon or Pinned). Under Mmap or Warm the engine
  looks at its own pointers and declines, so the switch would be one that silently does nothing. Off
  by default.

## A bug this found: a tied output head kept a pointer nobody could see

Enabling the flag failed the `moe_gates_gemma4` byte-identity gate with a segfault. The tiny gemma4
model has no `output.weight`, so llama.cpp builds the output head from the token embedding table and
the model carries a **second** `ggml_tensor` over the same file bytes, identically named. The capture
pass keyed its map by name, kept one of them, and `--dense-weights anon` / `ahwb` rebound that one:
the twin went on reading the mmap for the whole run.

That is a bug independent of this flag, and a bad one. On a model past RAM the twin is the output
projection, a tensor whose per-token cost rivals all the routed experts together, served by page
faults from flash under precisely the two policies that exist to prevent it. Gemma is in this case,
and so is any architecture that ties its embeddings. The first fix was a blacklist and it was the
wrong shape of answer; the capture now records every distinct leaf object, deduplicated by address.
Bytes are still read once and memory allocated once (the same 70 anon buffers on the gemma4 gate).

What that is worth in throughput is not measured, and the changelog says so. The mechanism is
certain, the size of it is not.

## What this corrects

`docs/roadmap.md` used to call the gap between the engine's effective rate and the drive's "duty
cycle, not bandwidth" and leave it unsized. That was the wrong diagnosis. The same entry, and the
two verdicts that lanes and threads are dead on the desktop host, are corrected in place rather than
quietly dropped.

`docs/architecture.md`, `AGENTS.md` and the README knob table are updated too. The README gains a row
for `--release-mmap` and one for `--row-stream`, which shipped earlier in 0.24.0 without ever
reaching the table where a user looks for a knob.

## Gates

`ctest` 13/13 green, including the byte-identity gates. clang-format 18 clean on all 17 changed
sources.

Desktop smoke test on the real model (Qwen3.6-35B-A3B Q4_K_M, 24 tokens, consecutive cells), which
is the check that matters because the host gates run on a synthetic model where this feature is
nearly inert:

| | baseline | `--release-mmap` |
|---|---:|---:|
| decode | 2.591 tok/s | **3.983 tok/s** |
| flash stall | 0.251 s/token | **0.101 s/token** |
| read per token | 271.52 MiB | 271.52 MiB |
| cache hit | 36.7% | 36.7% |
| evictions / re-reads | 6013 / 2746 | 6013 / 2746 |

Same bytes, same hit rate, same text, served in parallel. Also validated on the test phone through
the demo app.


Helldez:
Update: turning this on by default is not safe, and finding out why bounds what the feature can promise.

The `moe_gates_gemma4` byte-identity gate segfaults with the release enabled. A bisect (unmap the view but keep the section, then the reverse) put the fault on unmapping, and a vectored exception handler reported the faulting addresses as file offsets 12192 and 79264 — both inside `token_embd.weight`. The accounting said every gguf tensor was owned by a policy, and it was right: that model has no `output.weight`, so llama.cpp builds the output head from the embedding table and the model carries a **second tensor over the same mapped bytes**. The engine rebinds the tensor the gguf names; the tied twin is not a gguf tensor, so no name-based accounting can see it. The qwen3moe gate, which has a separate `output.weight`, passes.

That is a latent crash in the flag as first pushed, so it is fixed here rather than left for the default question: the session now declines when the gguf has no `output.weight`, which covers every architecture that ties its embeddings, Gemma among them. Verified both ways on the synthetic models, and Qwen3.6-35B still releases and still gains.

Two alternatives were measured while looking for a design that could not fail this way, and both are refuted:

- **Close the section handle, keep the view mapped.** Safe by construction, since no pointer can dangle — but it does not lift the serialisation (792 MiB/s against 790 with the mapping fully alive). It is the view, not the handle.
- **Read buffered while the mapping is alive.** Buffered reads do escape the serialisation, but the microbenchmark cannot price it honestly: `--buffered --mmap` measures 3406 MiB/s precisely because it is served from the page cache the mapping populated, which is the pollution O_DIRECT exists to avoid on a model many times larger than RAM. Settling it needs an engine-level A/B, not an iobench cell.

So this stays **opt-in**. What the session does is refuse every shape where a surviving pointer into the mapping is known — dense weights still mmap'd, an unowned file tensor under the MTP draft, a tied output head — and that is a blacklist, not a proof of safety. A positive proof needs a way to enumerate a loaded model's tensors, which llama.cpp does not expose today.

Helldez:
Update, superseding the comment above: the tied-head blacklist was the wrong shape of answer, and replacing it turned up a bug worth more than the flag it was blocking.

## The twin was not a special case, it was a weight we failed to rebind

When a gguf has no `output.weight`, llama.cpp builds the output head from the token embedding table, so the model holds **two `ggml_tensor` objects, identically named, over the same file bytes**. The capture pass kept weights in a map keyed by name, so it recorded one of them, and `--dense-weights anon` / `ahwb` rebound that one. The twin went on reading the model's mmap for the whole run.

That is not just a hazard for releasing the mapping. On a model past RAM it means the **output projection** — a tensor whose per-token cost rivals all the routed experts together — was being served by page faults from flash under the two policies that exist precisely to prevent that. Every architecture that ties its embeddings is in this case, Gemma among them.

So the capture now records every distinct leaf object, deduplicated by address, and a dense entry carries the other tensors over its range as aliases, rebound onto the same buffer. Bytes are read once and memory allocated once: the gemma4 gate reports the same 70 anon buffers as before.

What that is worth in throughput is **not measured**. The mechanism is certain, the size of it is not — it depends how much of the mapped table the kernel was still holding, which on a phone under reclaim is a different story from a desktop with room to spare. Gemma-4-26B on device is the cell that would price it, and it is owed.

## With no twin left behind, the decision stops being a guess

The release condition is no longer a list of shapes known to be unsafe. The engine asks the OS whether any weight the capture observed still points inside a mapping of the model files, and declines if any does (`pio::addresses_in_file_mappings`). One question covers every residency policy: a dense set left mmap'd, a table held back as oversized, a twin no name-based accounting could have found.

There is no model name, architecture name or tensor name anywhere in the change. Twins are folded by `(shard index, file offset)` — geometry of the file — and the release is decided from actual pointer addresses.

## Evidence, this host

| model | result | twins found | views released |
|---|---|---:|---:|
| tiny qwen3moe | releases | 0 | 1 |
| tiny gemma4 (tied head) | releases | **1** | 1 |
| tiny q4 | releases | 0 | 1 |
| tiny split, 4 shards | releases | 0 | **4** |
| Qwen3.6-35B-A3B Q4_K_M | releases | 0 | 1 |

The behaviour is discovered per model: one twin only on the tied-head model, four views only on the four-shard one.

Generated text with the flag and without: identical on all five, Qwen3.6-35B over 96 tokens included (572 characters, same SHA), while decode goes **2.72 → 3.93 tok/s**. The opposite direction is checked too: run `--dense-weights mmap` and the engine finds 39 weights still inside the mapping and stands down. All 13 byte-identity gates pass; CI is green on Windows, Linux, macOS and the Android APK.

## Still opt-in

The check answers for the pointers the capture pass observed and for no others: a graph shape a session never builds could hold another one, and llama.cpp exposes no way to enumerate a loaded model's tensors and settle it. That residue, plus the gguf tensors no policy owns once the MTP draft adds a second graph, is the reason the default stays off for now.

gjjkbssg:
I ran the missing macOS/APFS cell for the mmap-serialisation question here, as a post-merge measurement follow-up.

One important correction before the result: my first pass used 4 MiB reads, where the Mac baseline itself barely scaled (4L/1L ≈ 1.16×). That made the negative result non-identifiable, so I discarded that conclusion and first searched for request shapes with actual concurrency headroom.

On the internal APFS/NVMe volume:

- 576 KiB (#185's request shape): 1L 1065 MiB/s → 4L 1612 MiB/s (S4 = 1.51×)
- 256 KiB: 1L 921 → 4L 1609 MiB/s (S4 = 1.75×)

I then ran baseline / live-mmap / release+reopen at 4 lanes, 7 rotated interleaved rounds for each size. `bmoe-iobench` uses the same deterministic per-lane offset stream across cells, so the LBA trace is paired.

One macOS trap I hit during setup: a freshly normally-written test file was evidently still being served from resident data despite the read descriptors using `F_NOCACHE` — I saw ~12.3 GiB/s at one lane. I regenerated the file through an `F_NOCACHE` write path before taking the results below. I did not try to establish the exact Darwin cache-invalidation semantics here.

Median of the per-round live-mmap / baseline ratios:

- 576 KiB: **0.994**
- 256 KiB: **0.999**

Median of the per-round release+reopen / baseline ratios was 0.999 and 1.004 respectively.

So on this Mac/APFS configuration I don't reproduce the Windows-style mapping-induced throughput collapse: with baseline S4 of 1.51× and 1.75×, the live mapping preserves essentially all observed four-lane throughput. This rules out a Windows-scale effect under these conditions; it is not a sub-1% precision claim.

The 4 MiB cells are still useful as a negative-method example: that workload saturates around 1.3–1.6 GiB/s on this system, so a flat mapping knob there could not answer the serialisation question.

For #182, this leaves the previous result unchanged in scope: the +56% remains the measured effect of enabling `F_NOCACHE` **with the mapping live**. Whether that effect changes after releasing the mapping is a separate 2×2 interaction question that I have not measured.

I also haven't run an engine-level `--release-mmap` A/B on macOS, so the only prediction I'd make from these cells is little or no **direct read-throughput** gain from releasing the mapping at these request shapes; VM-footprint / memory-pressure effects remain open.

I have the raw interleaved runs locally if you want them curated into the bench-data set.

PR #197, opened 2026-09-16
feat(serve): OpenAI-compatible bridge and ARM64 Linux bundle (0.24.1)

cjl4hd:
## What

Three related pieces, one coherent change (0.24.1):

1. **Serve the engine to agent tooling.** The engine has no HTTP server — it serves prompts over the `--session` stdin protocol (the same protocol the Android app drives). `scripts/bmoe-serve.py` (stdlib-only) bridges that protocol into an OpenAI-compatible HTTP endpoint (`/v1/chat/completions` streaming and non-streaming, `/v1/models`, `/health`), keeping the model loaded and the expert cache warm between requests. `scripts/run-server.sh` resolves engine + model. No engine changes, no new dependencies, no llama.cpp contact — the bridge wraps the existing public protocol. Documented in the new `docs/serve.md`, README Quickstart, docs index, and telemetry.md cross-link.

2. **ARM64 GNU/Linux bundles.** `scripts/build-arm64.sh` cross-compiles and stages a self-contained `bmoe-arm64/` bundle (CLI + shared libs, RUNPATH `$ORIGIN/lib`, baseline `armv8.2-a+dotprod+fp16`, expert-ready hook included) so the server can run on an SBC or ARM box. Same CPU-baseline rationale as `build-android.ps1` (no i8mm); not the Android build.

3. **Fix: engine self-reported 0.23.0 after the 0.24.0 cut.** `project(VERSION)` is the single source of `BMOE_VERSION` (`--version`, the `engine=` preamble of every metrics file) and is meant to move with the app version in the release PR — the same regression as 0.20.0. Bumped to 0.24.1 with the app version (versionCode 40), changelog section dated in the same change, and the stray `</content>` line leaked at the end of CHANGELOG.md by the 0.24.0 commit removed.

## Verification

- Host gates: `ctest` 13/13 pass on the 0.24.1 build (byte-identity: qwen3moe, gemma4, split shards).
- Bridge smoke-tested end-to-end against `lfm2moe` (LFM2.5-8B-A1B Q4_K_M): streaming + non-streaming completions, `reasoning_content` separation, perf block, cancel-on-disconnect; ~8 tok/s warm at 96% cache hit.
- Compute-buffer trap documented: at ctx 8192 the wide graph reserved 4.1 GiB and OOM-killed the session on an 11 GiB host; `--ubatch 512` caps it at 258 MiB with decode unaffected (decode unaffected per the flag's contract).
- ARM64 build: configure reports the expert-ready hook present; artifact inspected (aarch64 ELF, `$ORIGIN/lib` RUNPATH resolves the five NEEDED libs).


Helldez:
Thanks! I'm reviewing it, trying to give you an answer tomorrow

cjl4hd:
This may be similar to #161, my goal was minimal changes.

Helldez:
Thanks for this, and sorry for the wait. I read all of it and ran the bridge locally.

The bridge I would rather not land. I have a native `bmoe-server` with the same endpoints in progress and I will open it as a draft soon, so the repo would end up with two ways to do the same thing. While testing yours I hit two bugs you will want to know about even for your own use:

1. If a client disconnects mid-stream, the next request comes back 200 with empty content, and every reply after that is shifted by one. The cancelled generation's remaining lines stay in the pipe, and the bridge never checks the `id` the events carry.
2. `"reset":1` lines get appended instead of replacing, so text the parser moves into reasoning stays in `content` as well.

I can share the small fake engine I used to reproduce both.

`build-arm64.sh` I would hold until the server is in: the bundle worth shipping is the one that carries it, and I have no ARM Linux box to check it on. That is where I would like your help. When the draft is up, would you try it with opencode on your ARM machine, and bring the script back extended to the server?

Your finding about strict SDKs rejecting SSE events without `choices` is going straight into the native server, thank you for that.

I will close this one once the draft is up, so you have something to look at first.

cjl4hd:
Sounds good, thanks for the review and pointing out the bugs. Since you're implementing bmoe-server, I'll pull your version once in. 

For build-arm64.sh, I'm still deciding between opencode and aider since aider is more token efficient, but regardless I'm happy to work on that once the draft is up. For context, I'm using build-arm64.sh for my cell phone running in proot ubuntu in termux. I cross compile from my laptop for convenience.

I do also plan to get some benchmarks up per your request in Issues once I'm happy with my setup. I have a few configurations that will take some time to tune. 

I also have a residency branch for pure transformer models that I can PR if you're interested. It won't help with hybrid models like LFM2.5 8BA1B, but for models like Qwen3-30B-A3B and Laguna XS 2.1 33BA3B, it avoids prefill with every turn. I'm asking here just to avoid overloading you with PRs.

Residency summary: An engine-side conversation cache, implemented entirely through llama.cpp's public API. On each request the engine renders its chat template over the incoming array, diffs it against what's already resident in KV, keeps the shared prefix, truncates the KV past the divergence point, and prefills only the new tail. Two-turn gate test: turn 1 prefilled 131 tokens, turn 2 prefilled 89 with 131 reused. On my laptop prefill is taking ~6min, and each new response is only seconds. Previously this meant 6 minutes each query, but now I'm only paying the 6 minutes on the first query.

